PR #4, opened 2022-05-19
Added password constraints

ivanfhtec:
Passwords must be at least 8 characters long, have at least one uppercase letter, lowercase letter, and one number.


DraganMitrasinovicHtec (commented):
Seems like a cleaner way to use modelMapper. Pattern is fine, latter we might do it in a seperate function.

PR #5, opened 2022-05-25
SJQ2T-16: Dockerised databases, cleaned up dependencies

ivanfhtec:
Docker compose sets up three services:
- one for MySQL
- one for Cassandra
- and a third service that waits for Cassandra to initialise and creates a keyspace (schema), if one doesn't exist

MySQL connects to port 3306 by default, so the dockerised version is visible through a different port (3305).

Should we also run maven from docker-compose?

### DIFF
--- a/backend/docker-compose.yml
+++ b/backend/docker-compose.yml
@@ -1,22 +1,22 @@
 services:
 
-  mysqldb:
-    image: mysql:8
-    container_name: mysqldb
+  skynest-db:
+    image: mysql:latest
+    container_name: skynest-db
     ports:
-      - 3305:3306
+      - 3306:3306
     restart: unless-stopped
     volumes:
       - db:/var/lib/mysql
     environment:
-      MYSQL_DATABASE: 'SkyNest'
+      MYSQL_DATABASE: 'skynest'
       MYSQL_USER: 'user'
       MYSQL_PASSWORD: '123456'
       MYSQL_ROOT_PASSWORD: '123456'
 
-  cassandra:
+  skynest-token-db:
     image: cassandra:latest
-    container_name: cassandra
+    container_name: skynest-token-db
     ports:
       - 9042:9042
     environment:
@@ -31,11 +31,11 @@ services:
       timeout: 10s
       retries: 10
 
-  cassandra-load-keyspace:
-    container_name: cassandra-load-keyspace
+  skynest-token-db-init:
+    container_name: skynest-token-db-init
     image: cassandra:latest
     depends_on:
-      cassandra:
+      skynest-token-db:
         condition: service_healthy
     volumes:
       - ./src/main/resources/cassandra_init_keyspace.cql:/schema.cql
--- a/backend/docker-compose.yml
+++ b/backend/docker-compose.yml
@@ -1,0 +1,45 @@
+services:
+
+  mysqldb:
+    image: mysql:8
+    container_name: mysqldb
+    ports:
+      - 3305:3306
+    restart: unless-stopped
+    volumes:
+      - db:/var/lib/mysql
+    environment:
+      MYSQL_DATABASE: 'SkyNest'
+      MYSQL_USER: 'user'
+      MYSQL_PASSWORD: '123456'
+      MYSQL_ROOT_PASSWORD: '123456'
+
+  cassandra:
+    image: cassandra:latest
+    container_name: cassandra
+    ports:
+      - 9042:9042
+    environment:
+      - "MAX_HEAP_SIZE=256M"
+      - "HEAP_NEWSIZE=128M"
+    restart: unless-stopped
+    volumes:
+      - ./out/cassandra_data:/var/lib/cassandra
+    healthcheck:
+      test: [ "CMD", "cqlsh", "-u cassandra", "-p cassandra" ,"-e describe keyspaces" ]
+      interval: 15s
+      timeout: 10s
+      retries: 10
+
+  cassandra-load-keyspace:
+    container_name: cassandra-load-keyspace
+    image: cassandra:latest
+    depends_on:
+      cassandra:
+        condition: service_healthy
+    volumes:
+      - ./src/main/resources/cassandra_init_keyspace.cql:/schema.cql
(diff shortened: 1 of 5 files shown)


backend/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+    restart: unless-stopped
+    volumes:
+      - db:/var/lib/mysql
+    environment:
+      MYSQL_DATABASE: 'SkyNest'
+      MYSQL_USER: 'user'
+      MYSQL_PASSWORD: '123456'
+      MYSQL_ROOT_PASSWORD: '123456'
+
+  cassandra:
```
petarstoimenovskihtec: Why would you use Cassandra? What is the reason behind it?

ivanfhtec: We'll need to store invalid tokens somewhere until they expire. A NoSQL database is probably a good choice.

Cassandra has an easy solution for this: any inserted row can have a time-to-live defined, so tokens can live in the database just until they expire.
I've also managed to make this work in code using the Cassandra implementation of CrudRepository, but there's a bit of a workaround involved.

I think MongoDB has something similar - Cassandra just seemed more straightforward in this regard.

petarstoimenovskihtec: Interesting, I would like to see this in action ;)

backend/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
+services:
+
+  mysqldb:
+    image: mysql:8
+    container_name: mysqldb
+    ports:
+      - 3305:3306
```
petarstoimenovskihtec: You can use the same port. Docker has its internal file system and networking so there won't be any interference.

backend/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+    restart: unless-stopped
+    volumes:
+      - db:/var/lib/mysql
+    environment:
+      MYSQL_DATABASE: 'SkyNest'
```
petarstoimenovskihtec: usually practice is to use small letters not camel case

backend/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
+services:
+
+  mysqldb:
+    image: mysql:8
+    container_name: mysqldb
```
petarstoimenovskihtec: for better readability name this skynestdb or something that will represent that this is tightly connected to SkyNest project. In future you may work on several projects and in that case you can get easily confused. Same applies for other services defined here

backend/docker-compose.yml
ivanfhtec: All of us on the backend have MySQL already installed, so we all have a service running on 3306.
Mapping Docker's internal port 3306 to 3305 seemed to be the easiest solution.

backend/src/main/resources/application.properties
```diff
@@ -1,8 +1,22 @@
-spring.datasource.url=jdbc:mysql://localhost:3306/SkyNest?createDatabaseIfNotExist=true
+
+#   MySQL
+spring.datasource.url=jdbc:mysql://localhost:3305/SkyNest?createDatabaseIfNotExist=true
 spring.datasource.username=user
 spring.datasource.password=123456
 spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.MySQL5InnoDBDialect
 spring.flyway.baseline-on-migrate=true
 #spring.jpa.hibernate.ddl-auto=none
```
petarstoimenovskihtec: Commented code should never be part of a commit, either uncomment it or remove it

backend/src/main/resources/application.properties
```diff
@@ -1,8 +1,22 @@
...
 #spring.jpa.show-sql: true
-#springdoc.swagger-ui.path=/swagger-ui.html
\ No newline at end of file
+
+#   Docs
+#springdoc.swagger-ui.path=/swagger-ui.html
+
+#   Cassandra
+spring.data.cassandra.contact-points=localhost
+spring.data.cassandra.port=9042
```
petarstoimenovskihtec, line 11: Since you are using values here and in docker file as well, maybe it is good option to consider extract those values in single file. If not with this PR, mark as TODO and create a technical dept task for it in backlog or something

backend/docker-compose.yml
petarstoimenovskihtec: Still, the app will communicate with Docker 3306 port, you via DB client will communicate with 3306 port from your local machine. The Docker 3306 port will communicate with 3306 from your local machine, it may be set as the same port in value, but it is not the same port logically.

petarstoimenovskihtec (commented):
Some small remarks

backend/docker-compose.yml
ivanfhtec: Agreed, will change.

backend/docker-compose.yml
ivanfhtec: Something like
mysqldb -> skynest-db
cassndra -> skynest-token-db
cassandra-load-keyspace -> skynest-token-db-load-keyspace

backend/src/main/resources/application.properties
ivanfhtec: Will remove.

backend/docker-compose.yml
petarstoimenovskihtec: sure, why not

backend/docker-compose.yml
ivanfhtec: Running it with
`    ports:
      - 3306:3306`
results in an error:
`Error response from daemon: driver failed programming external connectivity on endpoint mysqldb (cb1cad26117a0961d1fd3fc20cbd59443c13382711161a030546dd0c10c333fc): Error starting userland proxy: error while calling PortManager.AddPort(): listen tcp4 0.0.0.0:3306: bind: address already in use`

This is why I put 3305. Do I need to change something else as well for it to work with 3306?

ivanfhtec: I see the problem now, I'll change it to 3306. Thanks for the help.

backend/src/main/resources/application.properties
ivanfhtec: I've made it into a sub-task.
https://htecgroup.atlassian.net/browse/SJQ2T-55

markobarackovhtec (approved):

PR #8, opened 2022-05-26
Set up initial exception handling and error messages

DraganMitrasinovicHtec:
For user related errors, we will always throw UserException, which is caught by UserExceptionHandler.
ErrorMessage is the format passed to the frontend.
New types of user exceptions should be declared in UserExceptionType enum.

### DIFF
--- a/backend/src/main/java/com/htecgroup/SkyNest/exception/UserException.java
+++ b/backend/src/main/java/com/htecgroup/SkyNest/exception/UserException.java
@@ -1,0 +1,11 @@
+package com.htecgroup.SkyNest.exception;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@Getter
+@AllArgsConstructor
+public class UserException extends RuntimeException {
+
+  private UserExceptionType userExceptionType;
+}
--- a/backend/src/main/java/com/htecgroup/SkyNest/exception/UserException.java
+++ b/backend/src/main/java/com/htecgroup/SkyNest/exception/UserException.java
@@ -2,10 +2,17 @@ package com.htecgroup.SkyNest.exception;
 
 import lombok.AllArgsConstructor;
 import lombok.Getter;
+import org.springframework.http.HttpStatus;
 
 @Getter
 @AllArgsConstructor
 public class UserException extends RuntimeException {
 
-  private UserExceptionType userExceptionType;
+  private String message;
+  private HttpStatus status;
+
+  public UserException(UserExceptionType userExceptionType) {
+    message = userExceptionType.getMessage();
+    status = userExceptionType.getStatus();
+  }
 }
--- a/backend/src/main/java/com/htecgroup/SkyNest/exception/UserExceptionHandler.java
+++ b/backend/src/main/java/com/htecgroup/SkyNest/exception/UserExceptionHandler.java
@@ -1,0 +1,27 @@
+package com.htecgroup.SkyNest.exception;
+
+import com.htecgroup.SkyNest.model.response.ErrorMessage;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.ControllerAdvice;
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.context.request.WebRequest;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+@ControllerAdvice
+public class UserExceptionHandler {
+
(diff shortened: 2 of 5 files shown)


backend/src/main/java/com/htecgroup/SkyNest/exception/UserExceptionHandler.java
```diff
@@ -0,0 +1,27 @@
...
+import org.springframework.web.bind.annotation.ExceptionHandler;
+import org.springframework.web.context.request.WebRequest;
+
+import java.time.LocalDateTime;
+import java.time.format.DateTimeFormatter;
+
+@ControllerAdvice
+public class UserExceptionHandler {
+
+  private static final DateTimeFormatter formatter =
```
petarstoimenovskihtec: This is probably a good candidate for Util method/class.  Usually those kinds of variables are used in multiple places. You can create it to be more generic as well, for example to have a default option and one to propagate different pattern.

backend/src/main/java/com/htecgroup/SkyNest/exception/UserExceptionHandler.java
```diff
@@ -0,0 +1,27 @@
...
+      DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+
+  @ExceptionHandler(value = {UserException.class})
+  public ResponseEntity<ErrorMessage> handleUserException(UserException ex, WebRequest webRequest) {
+    ErrorMessage errorMessage =
+        new ErrorMessage(
+            ex.getUserExceptionType().getMessage(),
+            ex.getUserExceptionType().getStatus().value(),
+            LocalDateTime.now().format(formatter));
```
petarstoimenovskihtec: Also this `LocalDateTime.now().format(formatter)` can be extracted in a reusable and testable util method

backend/src/main/java/com/htecgroup/SkyNest/model/response/ErrorMessage.java
```diff
@@ -0,0 +1,14 @@
+package com.htecgroup.SkyNest.model.response;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.Setter;
+
+@Getter
+@Setter
```
petarstoimenovskihtec: do you need `@Setter ` ?

backend/src/main/java/com/htecgroup/SkyNest/exception/UserException.java
```diff
@@ -0,0 +1,11 @@
+package com.htecgroup.SkyNest.exception;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+
+@Getter
+@AllArgsConstructor
+public class UserException extends RuntimeException {
```
petarstoimenovskihtec, line 9: Maybe consider the option to generise this as well, since there will be different custom exceptions in this project, maybe add a parent abstract class and use that one in the ControllerAdvice class, it may save you lot of duplicate code. If you feel it is not the time for it, leave it for later with a TODO note

backend/src/main/java/com/htecgroup/SkyNest/model/response/ErrorMessage.java
DraganMitrasinovicHtec: I think not, because this class will only be used at the ControllerAdvice level. Will remove it.

backend/src/main/java/com/htecgroup/SkyNest/exception/UserExceptionHandler.java
DraganMitrasinovicHtec: You're right, it will look much cleaner too!

backend/src/main/java/com/htecgroup/SkyNest/exception/UserExceptionHandler.java
DraganMitrasinovicHtec: Agreed, will move it to util.

backend/src/main/java/com/htecgroup/SkyNest/exception/UserException.java
DraganMitrasinovicHtec: I think I get it, make it so we can throw an exception with a custom message / HttpStatus that might not make sense often, so it would be confusing to have it in the UserExceptionTypes. Will look into it

DraganMitrasinovicHtec (commented):
Read the comments from @petarstoimenovskihtec , will refine and refactor.

markobarackovhtec (approved):

PR #10, opened 2022-05-26
Creating and implementing initial style for login page

casualme2:
We've implemented Login page , the exact style the client wanted.
Also did small fixes (Actually very small fixes) to App and Loader components (removed return because the components , at the time, have only one child)

### DIFF
--- a/frontend/src/Components/Loader/Loader.jsx
+++ b/frontend/src/Components/Loader/Loader.jsx
@@ -3,9 +3,7 @@ import "./Loader.css";
 
 const Loader = () => (
 	<div className="container-fluid d-flex flex-column justify-content-center align-items-center vh-100">
-		<div className="spinner-border text-primary loader-circle-size" role="status">
-			<span className="sr-only">Loading...</span>
-		</div>
+		<div className="spinner-border text-primary loader-circle-size" role="status"></div>
 		<br />
 		<div className="text-muted loader-text-size">Loading your data...</div>
 	</div>
--- a/frontend/src/Components/Loader/Loader.jsx
+++ b/frontend/src/Components/Loader/Loader.jsx
@@ -1,16 +1,14 @@
 import React from "react";
 import "./Loader.css";
 
-const Loader = () => {
-	return (
-		<div className="container-fluid d-flex flex-column justify-content-center align-items-center vh-100">
-			<div className="spinner-border text-primary loader-circle-size" role="status">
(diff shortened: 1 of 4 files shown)


frontend/src/Components/Loader/Loader.jsx
```diff
@@ -1,16 +1,14 @@
...
-			<div className="text-muted loader-text-size">Loading your data...</div>
+const Loader = () => (
+	<div className="container-fluid d-flex flex-column justify-content-center align-items-center vh-100">
+		<div className="spinner-border text-primary loader-circle-size" role="status">
+			<span className="sr-only">Loading...</span>
 		</div>
-	);
-};
+		<br />
+		<div className="text-muted loader-text-size">Loading your data...</div>
```
marija-antic, line 8: Why having both Loading... and Loading your data...?
Also, break on new line can be done with CSS, but it's no big deal.

frontend/src/Components/Login/Login.jsx
```diff
@@ -1,6 +1,61 @@
 import React from "react";
 import "./Login.css";
+import logoImage from "./assets/logoblackandwhite.svg";
 
-export const Login = () => {
-  return <div>SkyNest</div>;
+const Login = () => {
+	return (
+		<section className="vh-100 container-fluid pageBackgroudColor">
```
marija-antic: This identation seems to big. I use Tabs 2 or Spaces 4

frontend/src/Components/Login/Login.css
```diff
@@ -0,0 +1,15 @@
+.imageLogo {
```
marija-antic: I thinnk I saw classes with `-` and now here with camelCase, just choose one style through project.

frontend/src/Components/Loader/Loader.jsx
casualme2: Fixed, thanks.

frontend/src/Components/Login/Login.css
casualme2: We wanted to separate Bootstrap styles with our custom style to know at any point, by just looking at it, if that's our custom style or bootstrap one.
But from now on we'll be consistent with naming them.

marija-antic: You can come up with a prefix on your custom classes like f.e. SkyNest -> sn-_class_name_. We do sth similar on projetc also.

PR #11, opened 2022-05-26
Implement login functionality for login and axios

casualme2:
Implemented Axios for Login page with MockAPI.
This is the one i used: https://github.com/nenadbugaric/interviews-reports-api-mock

Just wanted to have an endpoint where i can test different error response messages.


frontend/package.json
```diff
@@ -1,42 +1,42 @@
...
+	"version": "0.1.0",
+	"private": true,
+	"dependencies": {
+		"@testing-library/jest-dom": "^5.16.4",
+		"@testing-library/react": "^13.2.0",
+		"@testing-library/user-event": "^13.5.0",
+		"axios": "^0.27.2",
+		"bootstrap": "^5.1.3",
+		"react": "^18.1.0",
+		"react-bootstrap": "^2.4.0",
```
marija-antic, line 12: If you went with react-bootstrap, maybe you do not need scripts in html importing jquery and bootstrap from CDN?

frontend/src/Components/Login/Login.css
```diff
@@ -0,0 +1,15 @@
+.image-logo {
+	background-color: whitesmoke;
+}
+
+.little-image-logo {
+	width: 5rem;
+}
+
+.page-background-color {
+	background-color: #ebe7e8;
```
marija-antic: Define colors in one global file in variables and use it through the App. No need to do it in this PR but it's good practice and much cleaner to maintain.

frontend/src/Components/Login/Login.jsx
```diff
@@ -1,6 +1,117 @@
...
+
+	const emailRef = useRef();
+
+	const getUserToken = async () => {
+		AxiosInstance.post("/login", {
+			email: email,
+			password: password,
+		})
+			.then(({ data }) => {
+				setAccessToken(data.accessToken);
```
marija-antic: What if you do not get `data` property for any reason from BE? This will break. Make sure to check those stuff to be safe on FE. Same for `response` on catch.

marija-antic: Also, try async/await syntax, for me it's cleaner.

frontend/src/Components/Login/Login.jsx
```diff
@@ -1,6 +1,117 @@
...
+													required
+												/>
+												<label className="form-label" htmlFor="emailInput">
+													Email address
+												</label>
+											</div>
+											<div className="form-outline mb-4">
+												<input
+													type="password"
+													onChange={(e) => setPassword(e.target.value)}
```
marija-antic: Inline event handlers have an impact of performance since their reference change on each rerender. No big deal but just better practice most of the time.

frontend/src/Components/Login/Login.jsx
```diff
@@ -1,6 +1,117 @@
...
+		emailRef.current.focus();
+	}, []);
+
+	useEffect(() => {
+		setErrorMsg("");
+	}, [email, password]);
+
+	const onFormSubmit = (e) => {
+		e.preventDefault();
+		getUserToken();
```
marija-antic: Shouldn't you await this function since it is asynchronous? Also, try case when user clicks submit mutliple times what will happen? Maybe it's good to disable submit while waiitng for response from api. Just a thought.

frontend/src/Components/Login/Login.css
casualme2: Will do that, thanks for the suggestion.

frontend/src/Components/Login/Login.jsx
casualme2: Yeah it will , i totally missed that, will fix it, thanks :)

frontend/src/Components/Login/Login.jsx
casualme2: Already fixed, thanks.

frontend/src/Components/Login/Login.jsx
casualme2: Already fiexed the async and await, thanks for the heads up :) 

I know that the user can click multiple times, the goal is to fix that and add a "Loading button as well as disabled inputs" in one of the next tasks.

frontend/package.json
casualme2: I've had small issues with bootstrap library, will look into that more in depth :)

frontend/src/Components/Login/Login.jsx
```diff
@@ -1,6 +1,138 @@
...
+import AxiosInstance from "../axios/AxiosInstance";
 
-export const Login = () => {
-  return <div>SkyNest</div>;
+const Login = ({ setAccessToken }) => {
+	const [email, setEmail] = useState("");
+	const [password, setPassword] = useState("");
+	const [errorMsg, setErrorMsg] = useState("");
+	const [showPassword, setShowPassword] = useState(false);
+	const [loading, setLoading] = useState(false);
```
anticmarija, line 11: Multiple useState-s can be a good candidate for useReducer hook. Not a must, just a thought.

marija-antic (approved):

PR #12, opened 2022-05-27
SJQ2T-24 and SJQ2T-41: added registration db script, entities, dtos, and repositories

ivanfhtec:
also switched repos to use optional, where applicable
removed Utils as we'll be using UUIDs instead of generating our own strings


backend/src/main/java/com/htecgroup/SkyNest/controller/UserController.java
```diff
@@ -18,12 +17,10 @@ public class UserController {
   @PostMapping("/register")
   public UserResponse registerUser(@RequestBody UserRegisterRequest userRegisterRequest) {
 
-    ModelMapper modelMapper = new ModelMapper();
+    UserDto userDto =
+        userService.registerUser(new ModelMapper().map(userRegisterRequest, UserDto.class));
```
markobarackovhtec: Can we maybe create model mapper as bean in configuration, and to autowire that where we need it ?

backend/src/main/java/com/htecgroup/SkyNest/model/enitity/RoleEntity.java
```diff
@@ -0,0 +1,28 @@
...
+@AllArgsConstructor
+public class RoleEntity implements Serializable {
+
+  private static final long serialVersionUID = -6954026563601179057L;
+
+  public static final String ROLE_WORKER = "role_worker";
+  public static final String ROLE_MANAGER = "role_manager";
+  public static final String ROLE_ADMIN = "role_admin";
+
+  @Id @GeneratedValue private UUID id;
```
markobarackovhtec: I see that somewhere in a code you use annotation above some code, and somewhere inline, maybe is better to be consistent about this and for example to use it always above

backend/src/main/java/com/htecgroup/SkyNest/repository/RoleRepository.java
```diff
@@ -0,0 +1,13 @@
...
+
+import com.htecgroup.SkyNest.model.enitity.RoleEntity;
+import org.springframework.data.repository.CrudRepository;
+import org.springframework.stereotype.Repository;
+
+import java.util.Optional;
+import java.util.UUID;
+
+@Repository
+public interface RoleRepository extends CrudRepository<RoleEntity, UUID> {
```
markobarackovhtec: why somewhere CrudRepository and somewhere JpaRepository ?

backend/src/main/resources/db/migration/V2__registration_entities.sql
```diff
@@ -0,0 +1,69 @@
+--  modify old user entity
+ALTER TABLE `user`
```
markobarackovhtec: My suggestion here is to use multiple script for every alter, every create, every insert...

backend/src/main/java/com/htecgroup/SkyNest/controller/UserController.java
ivanfhtec: Ok, I've added a bean to SkyNestApplication.java and autowired it where necessary.

backend/src/main/java/com/htecgroup/SkyNest/model/enitity/RoleEntity.java
ivanfhtec: We're using Google's formatter, looks like it tries to fit all annotations in one line.

backend/src/main/java/com/htecgroup/SkyNest/repository/RoleRepository.java
ivanfhtec: JpaRepository extends CrudRepository, PagingAndSortingRepository, QueryByExampleExecutor, and Repository.

I thought that for roles, we don't need most of these. Should we have every repository extend JpaRepository?

backend/src/main/resources/db/migration/V2__registration_entities.sql
ivanfhtec: Is it enough to break the whole script into:
- modify user entity
- create role
- create company
- add foreign keys
- init role and company

In other words, break it up where I've left comments.

Or should I break it down further?

backend/src/main/java/com/htecgroup/SkyNest/repository/RoleRepository.java
markobarackovhtec: Valid answer, if you are sure that you are not going to use the functions provided by JpaRepository, use CrudRepository :)

backend/src/main/java/com/htecgroup/SkyNest/model/enitity/RoleEntity.java
markobarackovhtec: Then im ok with it

backend/src/main/resources/db/migration/V2__registration_entities.sql
markobarackovhtec: I would say script per alter, script per create, scrip per insert (you can use one insert for multiple values in same table)

ivanfhtec: Ok, I'll make one script per statement.

markobarackovhtec: In that case, scripts will have understandable names and we will know which script does what just by reading the name

markobarackovhtec:
Gj for using optional with suppliers.
Only thing that i dont see are unit tests :)

ivanfhtec:
I've added some tests for UserServiceImpl.

markobarackovhtec (approved):

markobarackovhtec:
Now i like it more :)

PR #13, opened 2022-05-27
SJQ2T-16 hotfix: cassandra keyspace is now successfully initialised

ivanfhtec:


DraganMitrasinovicHtec (approved):
Seems legit, works now.

PR #15, opened 2022-05-27
Sjq2t-17_authentication_be

DarkoVeljanoskiHtec:
Everything is set up for the authentication, I've added two new filters for the authentication and authorization and made changes to the web security class, register and login works.
The get users functionality is not implemented, I'll create new task on Jira for that.


backend/src/main/java/com/htecgroup/SkyNest/model/request/UserLoginRequest.java
```diff
@@ -1,8 +1,12 @@
 package com.htecgroup.SkyNest.model.request;
 
 import lombok.Data;
+import lombok.Getter;
+import lombok.Setter;
 
 @Data
+@Setter
```
DraganMitrasinovicHtec: Doesn't Data already make all Getters and Setters?

DarkoVeljanoskiHtec: I think you're right, I didn't know that.

backend/src/main/java/com/htecgroup/SkyNest/service/impl/UserServiceImpl.java
```diff
@@ -51,4 +51,6 @@ public UserDto findUserByEmail(String email) {
 
     return modelMapper.map(userEntity, UserDto.class);
   }
+
```
DraganMitrasinovicHtec: Try to format or check diffs before pushing, so we don't get changes like this :) .

DarkoVeljanoskiHtec: That's good suggestion, I'll check it from now.

backend/src/main/java/com/htecgroup/SkyNest/filter/CustomAuthenticationFilter.java
```diff
@@ -0,0 +1,65 @@
...
+                        .signWith(SignatureAlgorithm.HS512, SecurityConstants.TOKEN_SECRET)
+                        .compact();
+
+        UserService userService = (UserService) SpringApplicationContext.getBean("userServiceImpl");
+        UserDto userDto = userService.findUserByEmail(userName);
+
+        response.addHeader(SecurityConstants.HEADER_STRING, SecurityConstants.TOKEN_PREFIX + token);
+
+        response.addHeader("UserID", userDto.getId().toString());
+
```
DraganMitrasinovicHtec: Was it intended to have this line twice?

DarkoVeljanoskiHtec: it was mistake, I changed it.

PR #17, opened 2022-05-30
SJQ2T-25 validation be

MarkoRadunovicHtec:
Added UserValidator class to validate input fields, Added test class for UserValidator, added method to validate inputs in UserServiceImpl


backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -31,6 +33,8 @@ public UserDto registerUser(UserDto userDto) {
       throw new UserException(UserExceptionType.EMAIL_ALREADY_IN_USE);
     }
```
DraganMitrasinovicHtec, line 35: Should this maybe have a different name, it looks like it's supposed to return a boolean this way. Maybe something along the lines of validateUser()?

MarkoRadunovicHtec: You are right.. Will do it

backend/src/main/java/com/htecgroup/skynest/SkyNestApplication.java
```diff
@@ -22,4 +23,9 @@ public BCryptPasswordEncoder bCryptPasswordEncoder() {
   public ModelMapper modelMapper() {
     return new ModelMapper();
   }
+
+  @Bean
+  public UserValidator userValidator() {
```
petarstoimenovskihtec: Instead of defining the Bean here, you can use **@Component** annotation on Class level, in this case in UserValidator class.

MarkoRadunovicHtec: I'll change it.

PR #19, opened 2022-05-31
SJQ2T-17: Authentication BE

DarkoVeljanoskiHtec:
Refactored whole functionality about authentication, added authentication filter and updated Web Security accordingly.


backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -0,0 +1,73 @@
...
+
+    private final AuthenticationManager authenticationManager;
+    private final UserDetailsService userDetailsService;
+    private final PasswordEncoder passwordEncoder;
+    @Override
+    public Authentication attemptAuthentication(HttpServletRequest request, HttpServletResponse response) throws AuthenticationException {
+        UserLoginRequest credentials = new UserLoginRequest();
+        try {
+            credentials = new ObjectMapper().readValue(request.getInputStream(), UserLoginRequest.class);
+            log.info("Here, with help of object mapper we map the value of the request to UserLoginRequest class");
```
petarstoimenovskihtec: You don't need to explain what is done here. The purpose of log statement is to tell us what has been done in the app in case debugging is needed. 
Better log statement will be to say which user is trying to authenticate

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -0,0 +1,73 @@
...
+public class CustomAuthenticationFilter extends UsernamePasswordAuthenticationFilter {
+
+    private final AuthenticationManager authenticationManager;
+    private final UserDetailsService userDetailsService;
+    private final PasswordEncoder passwordEncoder;
+    @Override
+    public Authentication attemptAuthentication(HttpServletRequest request, HttpServletResponse response) throws AuthenticationException {
+        UserLoginRequest credentials = new UserLoginRequest();
+        try {
+            credentials = new ObjectMapper().readValue(request.getInputStream(), UserLoginRequest.class);
```
petarstoimenovskihtec: Since ObjectMapper probably will be used in different places, you can extract it as spring Bean and inject it here

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -0,0 +1,73 @@
...
+
+        if (!passwordEncoder.matches(credentials.getPassword(), userDetails.getPassword())) {
+            throw new UserException(UserExceptionType.PASSWORDS_DOES_NOT_MATCH);
+        }
+        return authenticationManager.authenticate(new UsernamePasswordAuthenticationToken(credentials.getEmail(), credentials.getPassword(), new ArrayList<>()));
+    }
+
+    @Override
+    protected void successfulAuthentication(HttpServletRequest request, HttpServletResponse response, FilterChain chain, Authentication authentication) throws IOException, ServletException {
+        log.info("We get in this method, only if we return from attemptAuthenticate method successfully");
```
petarstoimenovskihtec: Same here.. better log statement will be to say after userName var that that user is successfully logged in

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -0,0 +1,73 @@
...
+        String userName = ((User) authentication.getPrincipal()).getUsername();
+
+        String token = JWT.create()
```
petarstoimenovskihtec: This can be extracted in separate class and method. That way it can be reused if needed and tested as a unit, also is not tightly coupled

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -0,0 +1,73 @@
...
+                .withSubject(userName)
+                .withExpiresAt(new Date(System.currentTimeMillis() + SecurityConstants.EXPIRATION_TIME))
```
petarstoimenovskihtec: Try to use LocalDate and LocalDateTime from java 8 whenever possible

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -0,0 +1,73 @@
...
+                .sign(Algorithm.HMAC256(SecurityConstants.TOKEN_SECRET));
+
+        response.addHeader(SecurityConstants.HEADER_STRING, SecurityConstants.TOKEN_PREFIX + token);
+
+        log.info("We create the token for the user, with that specific username, and we add the token in the header with key Authentication");
```
petarstoimenovskihtec: Maybe just change it to.. Jwt token successfully created for user {username}

backend/src/main/java/com/htecgroup/skynest/security/SecurityConstants.java
```diff
@@ -0,0 +1,15 @@
...
+public class SecurityConstants {
+
+    private SecurityConstants(){
+        throw new IllegalStateException("Utility class");
+    }
+    public static final long EXPIRATION_TIME = 180000; // 30 minutes
+    public static final String TOKEN_PREFIX = "Bearer ";
+    public static final String HEADER_STRING = "Authorization";
+    public static final String SING_UP_URL = "/users/register";
+    public static final String TOKEN_SECRET = "jfgi95jgnfi4";
```
petarstoimenovskihtec, line 15: TOKEN_SECRET property should be part of .env file as secrets or even part of application.properties and just inject it here. That way for every env you can use different token_secret which is preferable. EXPIRATION_TIME as well.

backend/src/main/java/com/htecgroup/skynest/security/SecurityConstants.java
```diff
@@ -0,0 +1,15 @@
+package com.htecgroup.skynest.security;
+
+public class SecurityConstants {
+
+    private SecurityConstants(){
```
petarstoimenovskihtec: Since you already use Lombok, check how a private constructor can be done with it.

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
```diff
@@ -23,15 +26,23 @@ protected void configure(HttpSecurity http) throws Exception {
...
-        .antMatchers(HttpMethod.POST, "/users/register")
+        .antMatchers(HttpMethod.POST, SecurityConstants.SING_UP_URL)
         .permitAll()
-        //        .anyRequest()
-        //        .authenticated()
-        .and()
+        .anyRequest()
+        .authenticated()
+        .and().addFilter(getAuthenticationFilter())
+      //  .addFilter(new CustomAuthorizationFilter(authenticationManager()))   TODO: authorization filter
```
petarstoimenovskihtec: Remove commented out code

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
```diff
@@ -23,15 +26,23 @@ protected void configure(HttpSecurity http) throws Exception {
...
         .sessionManagement()
         .sessionCreationPolicy(SessionCreationPolicy.STATELESS);
   }
 
+
+  public CustomAuthenticationFilter getAuthenticationFilter() throws Exception {
+    final CustomAuthenticationFilter filter = new CustomAuthenticationFilter(authenticationManager(), userDetailsService, bCryptPasswordEncoder);
+    filter.setFilterProcessesUrl("/users/login");
```
petarstoimenovskihtec: Either use a constant file for endpoints or use it hardcoded. Currently you have both options :)

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
DarkoVeljanoskiHtec: Got it, I changed all the logs as well.

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -19,46 +19,48 @@
 @Service
 public class UserServiceImpl implements UserService {
 
-  @Autowired private UserRepository userRepository;
-  @Autowired private RoleRepository roleRepository;
-  @Autowired private BCryptPasswordEncoder bCryptPasswordEncoder;
-  @Autowired private ModelMapper modelMapper;
+    @Autowired private UserRepository userRepository;
```
petarstoimenovskihtec: you already used the approach with Lombok in CustomAuthenticationFilter.class. Use the same here instead of Autowired

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
DraganMitrasinovicHtec: Also, does ObjectMapper do the same thing as ModelMapper? If so we should probably use the same one everywhere.

DarkoVeljanoskiHtec: it doesn't work with ModelMapper, not sure why, so I've made ObjectMapper as a bean and inject it in the code.

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
DarkoVeljanoskiHtec: You mean the creation of the token?

backend/src/main/java/com/htecgroup/skynest/security/SecurityConstants.java
```diff
@@ -0,0 +1,15 @@
...
+        throw new IllegalStateException("Utility class");
+    }
+    public static final long EXPIRATION_TIME = 180000; // 30 minutes
+    public static final String TOKEN_PREFIX = "Bearer ";
+    public static final String HEADER_STRING = "Authorization";
+    public static final String SING_UP_URL = "/users/register";
```
ivanfhtec: just a spelling error, should be `SIGN_UP_URL`

backend/src/main/java/com/htecgroup/skynest/security/SecurityConstants.java
DarkoVeljanoskiHtec: I've change it, thanks.

backend/src/main/java/com/htecgroup/skynest/security/SecurityConstants.java
DarkoVeljanoskiHtec: If I'm not mistaken the .env file is still not created, so when that is done we will refactor this, I'll create Jira task as well.

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
DraganMitrasinovicHtec: I think that's what he meant, have a util class for making jwt tokens.

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
DarkoVeljanoskiHtec: Thanks, I've done it.

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
DarkoVeljanoskiHtec: I've added the login endpoint in the constants file.

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
DarkoVeljanoskiHtec: Thanks, I've done the same here.

backend/src/main/java/com/htecgroup/skynest/security/SecurityConstants.java
DarkoVeljanoskiHtec: Nice catch, I'll change it.

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
DarkoVeljanoskiHtec: I'm trying, but .withExpiresAt() requires argument with type Date, so not sure if I can change that.

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
DarkoVeljanoskiHtec: I've created the jwt util class, so hopefully that's what Petar meant.

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -0,0 +1,77 @@
...
+
+@AllArgsConstructor
+@Log4j2
+public class CustomAuthenticationFilter extends UsernamePasswordAuthenticationFilter {
+
+    private final AuthenticationManager authenticationManager;
+    private final UserDetailsService userDetailsService;
+    private final PasswordEncoder passwordEncoder;
+
+    private ObjectMapper objectMapper;
```
petarstoimenovskihtec: Mark as final

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -0,0 +1,77 @@
...
+
+    @Override
+    public Authentication attemptAuthentication(HttpServletRequest request, HttpServletResponse response) throws AuthenticationException {
+        UserLoginRequest credentials = new UserLoginRequest();
+        objectMapper = new ObjectMapper();
```
petarstoimenovskihtec: Since objectMapper is defined as bean, there is no need to be initialized here, it will already come initialized from spring. Try it in debug mode

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -0,0 +1,77 @@
...
+        try {
+            credentials = objectMapper.readValue(request.getInputStream(), UserLoginRequest.class);
+
+        } catch (IOException e) {
+            log.error(e);
```
petarstoimenovskihtec: Log the error with some meaningful message + exception thrown.. You can check the other PRs

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+        UserDetails userDetails = userDetailsService.loadUserByUsername(credentials.getEmail());
+        log.info("The user: " + credentials.getEmail() + " is trying to authenticate.");
```
petarstoimenovskihtec: Instead of string concatenation you can use {} as a placeholder in log.info and provide the param. Take a look how is done

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -0,0 +1,77 @@
...
+        }
+        return authenticationManager.authenticate(new UsernamePasswordAuthenticationToken(credentials.getEmail(), credentials.getPassword(), new ArrayList<>()));
+    }
+
+    @Override
+    protected void successfulAuthentication(HttpServletRequest request, HttpServletResponse response, FilterChain chain, Authentication authentication) throws IOException, ServletException {
+        String userName = ((User) authentication.getPrincipal()).getUsername();
+        log.info(userName + " is successfully logged in.");
+
+        Date expiration = new Date(System.currentTimeMillis() + SecurityConstants.EXPIRATION_TIME);
```
petarstoimenovskihtec, line 68: Take a look how this can be done with LocalDateTime

petarstoimenovskihtec (commented):
Just small remarks, when fixed, you can merge

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
DarkoVeljanoskiHtec: You're right, I marked it as final and it solved this as well.

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -9,56 +9,59 @@
...
-                () ->
-                    new UserException(
-                        "Role " + RoleEntity.ROLE_WORKER + " not found.",
-                        HttpStatus.INTERNAL_SERVER_ERROR));
-    userDto.setRole(modelMapper.map(roleEntity, RoleDto.class));
+        if (userRepository.existsByEmail(userDto.getEmail())) {
+            throw new UserException(UserExceptionType.EMAIL_ALREADY_IN_USE);
+        }
+
+        RoleEntity roleEntity =
```
petarstoimenovskihtec: I would create **_RoleService_** and add this **_findByName_** method there. That method will return in your case **_RoleDto_** object, with input attribute the role that you search, meaning that even the modelMapper should be there. This way method will be reusable everywhere where role needs to be returned based on role name.

petarstoimenovskihtec: I understand that this was done previously and it is not change in this PR , but I saw it now :)

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
petarstoimenovskihtec: ObjectMapper is used when you convert json to object and vice versa, ModelMapper is used when you map from one object type into another

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
petarstoimenovskihtec: hmm, ok probably with newer versions of jwt lib, that is changed (if there are any new versions :D ). Anyway it is ok for now.

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
DarkoVeljanoskiHtec: Actually, it isn't my code at all, but I'll try to solve it.

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -0,0 +1,78 @@
...
+        UserDetails userDetails = userDetailsService.loadUserByUsername(credentials.getEmail());
+        log.info("The user: {} is trying to authenticate.", credentials.getEmail());
+        if (!credentials.getEmail().equals(userDetails.getUsername())){
+            throw new UserException(UserExceptionType.INVALID_EMAIL_OR_PASSWORD);
+        }
+
+        if (!passwordEncoder.matches(credentials.getPassword(), userDetails.getPassword())) {
+            throw new UserException(UserExceptionType.PASSWORDS_DOES_NOT_MATCH);
+        }
+        return authenticationManager.authenticate(new UsernamePasswordAuthenticationToken(credentials.getEmail(), credentials.getPassword(), new ArrayList<>()));
```
petarstoimenovskihtec, line 59: Either provide list of roles instead of empty array or use the other constructor (I believe there is one constructor with 2 params)

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
petarstoimenovskihtec: Well actually you all as a team are responsible for the code that every each one of you write . That is one of the reasons why PRs exist in first place, so every each one of you can take a look what is happening, understand and jump in and fix whatever is necessary whenever is necessary @ivanfhtec @DraganMitrasinovicHtec @MarkoRadunovicHtec @MetoMladenovski

backend/src/main/java/com/htecgroup/skynest/security/SecurityConstants.java
petarstoimenovskihtec: that would be a nice functionality - SING UP :D

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -8,57 +8,55 @@
...
-  @Autowired private UserRepository userRepository;
-  @Autowired private RoleRepository roleRepository;
-  @Autowired private BCryptPasswordEncoder bCryptPasswordEncoder;
-  @Autowired private ModelMapper modelMapper;
+     private UserRepository userRepository;
+     private RoleRepository roleRepository;
 
-  @Override
-  public UserDto registerUser(UserDto userDto) {
+     private RoleService roleService;
```
petarstoimenovskihtec: RoleService needs to be part of Spring, so you need to inject in UserServiceImpl not provide as input param

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -8,57 +8,55 @@
...
 @Service
+@AllArgsConstructor
 public class UserServiceImpl implements UserService {
 
-  @Autowired private UserRepository userRepository;
-  @Autowired private RoleRepository roleRepository;
-  @Autowired private BCryptPasswordEncoder bCryptPasswordEncoder;
-  @Autowired private ModelMapper modelMapper;
+     private UserRepository userRepository;
+     private RoleRepository roleRepository;
```
petarstoimenovskihtec: Do you need this now?

backend/src/main/java/com/htecgroup/skynest/service/impl/RoleServiceImpl.java
```diff
@@ -0,0 +1,26 @@
...
+import org.springframework.http.HttpStatus;
+import org.springframework.stereotype.Service;
+
+@Service
+@AllArgsConstructor
+public class RoleServiceImpl implements RoleService {
+
+    private RoleRepository roleRepository;
+    @Override
+    public RoleEntity findByName(String name) {
```
petarstoimenovskihtec: Service should not return Entity. Meaning the mapper needs to be here as well

petarstoimenovskihtec: Check UserService, it returns Dto

petarstoimenovskihtec (approved):

PR #20, opened 2022-05-31
SJQ2T-55: Dockerise BE app

ivanfhtec:
The backend can now run from docker completely. This PR does not include the necessary `.env` file.

Flyway now also runs from docker since there's no reason for it to run every time the app starts.

The app can still run locally, just make sure to docker compose `skynest-db-flyway` and `skynest-token-db-init` which fire up the databases. After these services are finished, you can run the app (locally).


petarstoimenovskihtec:
The point of flyway is to run every time, not on demand.. 
Flyway stores in database what is the state that is migrated and what needs to be added additionally with every server start. If those 2 states are the same, then nothing will be triggered.

docker-compose.yml
```diff
@@ -0,0 +1,76 @@
...
+    container_name: skynest-db
+    ports:
+      - $DB_PORT_HOST:$DB_PORT_CONTAINER
+    restart: unless-stopped
+    env_file:
+      - db.env
+    volumes:
+      - db:/var/lib/mysql
+
+  skynest-db-flyway:
```
petarstoimenovskihtec: This one is not needed at all. Pls remove it and let flyway run everytime via application start

petarstoimenovskihtec (changes requested):
Pls see Flyway comment

ivanfhtec:
Flyway runs from spring again.

To reset the databases, delete the `.db_data/` folder (while the services are stopped).
Cassandra's cache used to be in a folder called `out/`, so feel free to delete.

petarstoimenovskihtec:
> Flyway runs from spring again.
> 
> To reset the databases, delete the `.db_data/` folder (while the services are stopped). Cassandra's cache used to be in a folder called `out/`, so feel free to delete.

You can simple remove the container, and recreate it again (docker rm <container_name>). With that all new changes will be automatically created.

README.md
```diff
@@ -1 +1,9 @@
-# SkyNest
\ No newline at end of file
+# SkyNest
+The Sky Nest cloud service is a scalable and safe system designed to provide data storage and user control.
+
+## Frontend
+
+## Backend
+Run backend server/databases using <code>docker compose up</code> in the root folder. A <code>.env</code> should also be in root. 
+
```
petarstoimenovskihtec, line 27: Good practice is to comment the whole process how the app is starting (for new colleagues, FE/QA team). For example add prerequisites section like Java 11, Docker etc etc.. How to run locally in IntelliJ (with all configurations , commands or what ever is needed) or via terminal if needed, what needs to be started before (like databases, commands like you already have done )

ivanfhtec:
> > Flyway runs from spring again.
> > To reset the databases, delete the `.db_data/` folder (while the services are stopped). Cassandra's cache used to be in a folder called `out/`, so feel free to delete.
> 
> You can simple remove the container, and recreate it again (docker rm <container_name>). With that all new changes will be automatically created.

So `docker rm skynest-db` will also remove `./.db_data/mysql`?

README.md
ivanfhtec: Ok, I'll expand the readme.

backend/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM maven:3.8.5-openjdk-11
+WORKDIR /backend
+COPY . .
+RUN mvn clean install -Dmaven.test.skip=true
```
petarstoimenovskihtec, line 4: Just a NOTE:  Create a task for maven profiles , to run this tests or not by separate different profiles and provide the profile here with -p <test_name>.. Don't change it now, just add a task as a Technical Dept

ivanfhtec: I've made it into a task.
[SJQ2T-58](https://htecgroup.atlassian.net/browse/SJQ2T-58)

I had to skip tests here since some tests don't really know they're inside a docker container and keep ignoring environment variables from `.env`.

backend/src/main/resources/application.properties
```diff
@@ -1,17 +1,30 @@
+#   local dev values
+DB_HOST=localhost
+DB_PORT=3306
+DB_NAME=skynest
+DB_USERNAME=user
+DB_PASSWORD=123456
```
petarstoimenovskihtec, line 6: Good approach for this variables and values is to have separate application.properties for different Spring profiles :) The files would be application-{profile}.properties where you set just the values for keys that are overridden in corresponding profile. Again create a technical dept task in backlog, you don't need to change it now

db.env
```diff
@@ -0,0 +1,4 @@
+MYSQL_DATABASE=${DB_NAME}
+MYSQL_USER=${DB_USERNAME}
```
petarstoimenovskihtec, line 2: Since this files are not commited, good approach is to add corresponding example file. In this case add _**db.env.example**_ file and that would have the same keys, just mocked (not correct) values. That way all other team members will have knowledge what needs to be added in their .env files

backend/src/main/resources/application.properties
ivanfhtec: It's a task now [SJQ2T-59](https://htecgroup.atlassian.net/browse/SJQ2T-59).

db.env
ivanfhtec: I thought I could just send the `.env` file to them. But there should still be some examples anyway?

petarstoimenovskihtec: correct.. the app needs to have the key examples but actually values you need to sent directly

ivanfhtec: The only file i didn't commit is `.env`. That's were all the values used in the other .env files will come from. Would it be enough to just make an `.env.example`?

petarstoimenovskihtec (approved):

PR #21, opened 2022-06-01
SJQ2T-25: Validation BE

MarkoRadunovicHtec:
Validation implemented in UserRegisterRequest, adequate handler added into CustomExceptionHandler, phone number already in use validation implemented

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/exception/UserExceptionHandler.java
+++ b/backend/src/main/java/com/htecgroup/skynest/exception/UserExceptionHandler.java
@@ -2,12 +2,19 @@ package com.htecgroup.skynest.exception;
 
 import com.htecgroup.skynest.model.response.ErrorMessage;
 import com.htecgroup.skynest.util.DateTimeUtil;
+import lombok.extern.slf4j.Slf4j;
+import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
+import org.springframework.validation.FieldError;
+import org.springframework.web.bind.MethodArgumentNotValidException;
 import org.springframework.web.bind.annotation.ControllerAdvice;
 import org.springframework.web.bind.annotation.ExceptionHandler;
 import org.springframework.web.context.request.WebRequest;
+import java.util.List;
+import java.util.stream.Collectors;
 
 @ControllerAdvice
+@Slf4j
 public class UserExceptionHandler {
 
   @ExceptionHandler(value = {UserException.class})
@@ -15,7 +22,24 @@ public class UserExceptionHandler {
     ErrorMessage errorMessage =
         new ErrorMessage(
             ex.getMessage(), ex.getStatus().value(), DateTimeUtil.currentTimeFormatted());
-
+    log.error("Handle UserException. Error: ", ex);
     return new ResponseEntity<>(errorMessage, ex.getStatus());
+  }
+
+  @ExceptionHandler(value = {MethodArgumentNotValidException.class})
+  public ResponseEntity<ErrorMessage> handleValidationException(
+      MethodArgumentNotValidException ex) {
+    List<String> errors =
+        ex.getBindingResult().getAllErrors().stream()
+            .map(
+                error ->
+                    String.format(
+                        "%s %s", ((FieldError) error).getField(), error.getDefaultMessage()))
+            .collect(Collectors.toList());
+    ErrorMessage errorMessage =
+        new ErrorMessage(
+            errors, HttpStatus.BAD_REQUEST.value(), DateTimeUtil.currentTimeFormatted());
+    log.error("Handle MethodArgumentNotValidException. Error: ", ex);
+    return new ResponseEntity<>(errorMessage, HttpStatus.BAD_REQUEST);
   }
 }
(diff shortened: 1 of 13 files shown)


backend/src/main/java/com/htecgroup/skynest/exception/CustomException.java
```diff
@@ -6,13 +6,13 @@
 
 @Getter
 @AllArgsConstructor
-public class UserException extends RuntimeException {
+public class CustomException extends RuntimeException {
```
petarstoimenovskihtec: Because lots of PR already use the UserException, it is good practice this refactoring to be done in separate PR

backend/src/main/java/com/htecgroup/skynest/exception/CustomExceptionHandler.java
```diff
@@ -0,0 +1,47 @@
...
+  @ExceptionHandler(value = {MethodArgumentNotValidException.class})
+  public ResponseEntity<ErrorMessage> handleValidationException(
+      MethodArgumentNotValidException ex) {
+    List<String> errors = new ArrayList<>();
+    ex.getBindingResult()
+        .getAllErrors()
+        .forEach(
+            error -> {
+              String errorMessage =
+                  ((FieldError) error).getField() + " " + error.getDefaultMessage();
```
petarstoimenovskihtec: you can use String.format instead of this concatenation

backend/src/main/java/com/htecgroup/skynest/model/request/UserRegisterRequest.java
```diff
@@ -1,14 +1,37 @@
...
+  @Email(message = "format is not valid", regexp = "(.+)@(\\S+)\\.(\\S+)")
+  @Size(max = 254, message = "length cannot be over 254 characters")
   private String email;
+
+  @Pattern(
+      regexp = "(?=.*[a-z])(?=.*[A-Z])(?=.*\\d)[a-zA-Z\\d]{8,50}",
+      message = "format not valid")
   private String password;
+
+  @NotBlank(message = "cannot be empty")
```
petarstoimenovskihtec: Can it be Null ? Applies to all required properties

backend/src/main/java/com/htecgroup/skynest/exception/CustomExceptionHandler.java
MarkoRadunovicHtec: Will do it.

backend/src/main/java/com/htecgroup/skynest/exception/CustomException.java
MarkoRadunovicHtec: Of corse. I ll redo changes. And than after it can be done in separate PR

backend/src/main/java/com/htecgroup/skynest/model/request/UserRegisterRequest.java
MarkoRadunovicHtec: @NotBlank forbid null values. So I can add that in message. For example Cannot be null or empty.

backend/src/main/java/com/htecgroup/skynest/exception/UserExceptionHandler.java
```diff
@@ -2,20 +2,45 @@
...
+@Slf4j
 public class UserExceptionHandler {
 
   @ExceptionHandler(value = {UserException.class})
   public ResponseEntity<ErrorMessage> handleUserException(UserException ex, WebRequest webRequest) {
     ErrorMessage errorMessage =
         new ErrorMessage(
             ex.getMessage(), ex.getStatus().value(), DateTimeUtil.currentTimeFormatted());
-
+    log.error(ex.toString(), ex);
```
petarstoimenovskihtec: it is better to use log.error("Handle UserException. Error reason: ", ex) or something like that. The way that you used you would log the same error twice I believe

backend/src/main/java/com/htecgroup/skynest/exception/UserExceptionHandler.java
```diff
@@ -2,20 +2,45 @@
...
+  public ResponseEntity<ErrorMessage> handleValidationException(
+      MethodArgumentNotValidException ex) {
+    List<String> errors = new ArrayList<>();
+    ex.getBindingResult()
+        .getAllErrors()
+        .forEach(
+            error ->
+                errors.add(
+                    String.format(
+                        "%s %s", ((FieldError) error).getField(), error.getDefaultMessage())));
```
petarstoimenovskihtec: Instead of forEach, take a look at **_map_** and _**collect**_ options and use them as combination.

backend/src/main/java/com/htecgroup/skynest/exception/UserExceptionHandler.java
MarkoRadunovicHtec: I willl fix it

backend/src/main/java/com/htecgroup/skynest/exception/UserExceptionHandler.java
MarkoRadunovicHtec: I'll take a look and use it.

petarstoimenovskihtec (approved):

MarkoRadunovicHtec:
Validation for User register request implemented, Handler added for MethodArgumentNotValidException, ErrorMessage structure edited

PR #22, opened 2022-06-01
SJQ2T-26: Email Verification BE

MetoMladenovski:
Implemented Email Verification. Made changes to register method to send mail to user on every register that contains a link with jwt, on which the user clicks and verifies his email. 

There was some kind of issue with .gitignore and the /out folder so some files from there are showing up in the commits.. I don't know if that is going to be a problem.


backend/out/cassandra_data/data/system_auth/roles-5bc52802de2535edaeab188eecebb090/nb-1-big-TOC.txt
```diff
@@ -0,0 +1,8 @@
+Index.db
+Statistics.db
+TOC.txt
```
DraganMitrasinovicHtec: Shouldnt't the out cassandra file be ignored by git?

MetoMladenovski: It should, but for some reason, unknown to me, it didn't.

MetoMladenovski:
I just remembered that there is no way user to verify himself if he misses the expiration time on the first mail(token).. So I am going to make some changes real fast.

backend/pom.xml
```diff
@@ -80,6 +80,21 @@
             <version>1.6.8</version>
         </dependency>
 
+        <!-- https://mvnrepository.com/artifact/io.jsonwebtoken/jjwt -->
+        <dependency>
+            <groupId>io.jsonwebtoken</groupId>
+            <artifactId>jjwt</artifactId>
+            <version>0.9.1</version>
+        </dependency>
```
MarkoRadunovicHtec: That dependancy for jwt have vulnerabilities listed in mvnrepository, I think it's better to use this dependency: <groupId>com.auth0</groupId>
<artifactId>java-jwt</artifactId>
<version>3.19.2</version>

DarkoVeljanoskiHtec: Yes, I used the same one in my current PR.

MetoMladenovski: I didn't see the vulnerabilities.. Will change it

markobarackovhtec:
Cassandra files should not be part of commit, and should not go up on origin

backend/out/cassandra_data/data/system_auth/roles-5bc52802de2535edaeab188eecebb090/nb-1-big-TOC.txt
petarstoimenovskihtec: Than explicitly remove those /out files and update the PR

backend/src/main/java/com/htecgroup/SkyNest/service/impl/EmailServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+import org.springframework.mail.javamail.MimeMessageHelper;
+import org.springframework.scheduling.annotation.Async;
+import org.springframework.stereotype.Service;
+
+import javax.mail.MessagingException;
+import javax.mail.internet.MimeMessage;
+
+@Service
+public class EmailServiceImpl implements EmailService {
+  @Autowired private JavaMailSender javaMailSender;
```
petarstoimenovskihtec: Pls discuss internally the two approaches, using @Autowire (not advisable :) ) or using controller injection using Lombok. I saw few different PRs with different approaches.. it is strongly advisable to have same code standard,
@DarkoVeljanoskiHtec @MarkoRadunovicHtec @ivanfhtec @DraganMitrasinovicHtec same applies for you all

backend/src/main/java/com/htecgroup/SkyNest/service/impl/EmailServiceImpl.java
```diff
@@ -0,0 +1,35 @@
...
+  private final Logger LOGGER = LoggerFactory.getLogger(EmailServiceImpl.class);
+
+  @Override
+  @Async
+  public void send(String to, String email) {
+    try {
+      MimeMessage mimeMessage = javaMailSender.createMimeMessage();
+      MimeMessageHelper helper = new MimeMessageHelper(mimeMessage, "utf-8");
+      helper.setText(email, true);
```
petarstoimenovskihtec: I would try to make this method a reusable one, cause maybe there would be different mails that needs to be sent and the code is pretty much the same, except to, email and subject

backend/src/main/java/com/htecgroup/SkyNest/service/impl/UserServiceImpl.java
```diff
@@ -61,4 +77,21 @@ public UserDto findUserByEmail(String email) {
...
+  public Boolean confirmEmail(String token) {
+    boolean validateToken = jwtEmailVerificationUtils.validateJwtToken(token);
+    if (validateToken) {
+      String email = jwtEmailVerificationUtils.getEmailFromJwtEmailVerificationToken(token);
+      UserEntity userEntity =
+              userRepository
+                      .findUserByEmail(email)
+                      .orElseThrow(
+                              () -> new UsernameNotFoundException("could not find user with email: " + email));
+      userEntity.setEnabled(true);
```
petarstoimenovskihtec: Usually all object should be immutable, meaning that if you are setting additional properties it needs to go in separate method, and return new object with all previous properties and the new ones.

backend/src/main/java/com/htecgroup/SkyNest/service/impl/UserServiceImpl.java
```diff
@@ -61,4 +77,21 @@ public UserDto findUserByEmail(String email) {
...
+      userEntity.setVerified(true);
+      userRepository.save(userEntity);
+      return userEntity.getEnabled() && userEntity.getVerified();
```
petarstoimenovskihtec: Due to the previous setters, isn't the return value always true?

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -13,16 +13,25 @@
 public class UserController {
 
   @Autowired private UserService userService;
-  @Autowired private ModelMapper modelMapper;
 
   @PostMapping("/register")
   public UserResponse registerUser(@RequestBody UserRegisterRequest userRegisterRequest) {
 
-    UserDto userDto = userService.registerUser(modelMapper.map(userRegisterRequest, UserDto.class));
+    ModelMapper modelMapper = new ModelMapper();
```
petarstoimenovskihtec: Would it be better and possible if the ModelMapper is part of sprint and inject it here? Pls check if that is done in some other PRs

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -13,16 +13,25 @@
...
+    userDto = userService.registerUser(userDto);
+
+    // for testing
 
     return modelMapper.map(userDto, UserResponse.class);
   }
 
+  @GetMapping("/confirm")
+  public Boolean confirmEmail(@RequestParam String token) {
+    return userService.confirmEmail(token);
```
petarstoimenovskihtec: Log statement would be useful here

backend/src/main/java/com/htecgroup/SkyNest/service/EmailService.java
```diff
@@ -0,0 +1,5 @@
+package com.htecgroup.skynest.service;
+
+public interface EmailService {
```
petarstoimenovskihtec: Btw I just notice.. SkyNest is deprecated package so needs to be removed , so comments are all applicable for new files

petarstoimenovskihtec (changes requested):
remove old packages and /out folder, and update PR

backend/src/main/java/com/htecgroup/SkyNest/service/impl/EmailServiceImpl.java
MetoMladenovski: If you are thinking about sending the subject as a parameter as well, I did that. I am not sure that i understood correctly.

backend/src/main/java/com/htecgroup/SkyNest/service/impl/UserServiceImpl.java
MetoMladenovski: I always like to return something to be sure that it's working during testing and trying it out.. 
I can maybe return ResponseEntity of a String.

backend/src/main/java/com/htecgroup/SkyNest/service/impl/UserServiceImpl.java
MetoMladenovski: I did the enabling of the user in another method enableUser.

backend/src/main/java/com/htecgroup/skynest/util/JwtEmailVerificationUtils.java
```diff
@@ -0,0 +1,137 @@
...
+    return JWT.require(Algorithm.HMAC256(jwtEmailVerificationSecret))
+        .build()
+        .verify(token)
+        .getSubject();
+  }
+
+  public boolean validateJwtToken(String token) {
+    try {
+      Date expiresAt =
+          JWT.require(Algorithm.HMAC256(jwtEmailVerificationSecret))
```
petarstoimenovskihtec: You can extract this in variable: JWT.require(Algorithm.HMAC256(jwtEmailVerificationSecret)) and reuse it

backend/src/main/java/com/htecgroup/skynest/util/JwtEmailVerificationUtils.java
```diff
@@ -0,0 +1,137 @@
...
+  private int jwtEmailVerificationExpirationMs;
+
+  @Value("${backend.app.confirmationLink}")
+  private String confirmationLink;
+
+  public String generateJwtEmailVerificationToken(String email) {
+    return JWT.create()
+        .withSubject(email)
+        .withIssuedAt(new Date())
+        .withExpiresAt(new Date((new Date()).getTime() + jwtEmailVerificationExpirationMs))
```
petarstoimenovskihtec: take a look at LocalDateTime, and try to use .addTime method

backend/src/main/java/com/htecgroup/SkyNest/service/impl/EmailServiceImpl.java
petarstoimenovskihtec: yes, you can even create an object that will contain every parameter needed for sending different mail. Fields such as: to, from, text, html, subject etc etc

backend/src/main/java/com/htecgroup/SkyNest/service/impl/UserServiceImpl.java
petarstoimenovskihtec: for testing you can always use log statements, just to double check the result and use debugging extensively. When you push code it means that this will be the version that everyone will use, so there won't be additional changes to remove the code that is just for testing purpose.

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -23,6 +27,21 @@ public UserResponse registerUser(@RequestBody UserRegisterRequest userRegisterRe
...
+    String response = userService.confirmEmail(token);
+    log.info("User successfully confirmed");
+    return ResponseEntity.ok(response);
+  }
+
+  @PostMapping("/resendEmail")
+  public ResponseEntity<String> resendUserEmail(@RequestParam String email){
+    userService.sendVerificationEmail(email);
+    String response = "Email resent successfully";
+    log.info("Email resend successfully");
```
petarstoimenovskihtec: This is the same message, you know the drill.. extract into variable :)

backend/src/main/java/com/htecgroup/skynest/service/EmailService.java
```diff
@@ -0,0 +1,5 @@
+package com.htecgroup.skynest.service;
+
+public interface EmailService {
+  void send(String to, String email, String subject);
```
petarstoimenovskihtec: As I wrote earlier, you can create a new object for email metadata and sent that as input param. For every different mail, that object will contain different data, but the same logic will be applied for sending mail .

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -42,17 +47,32 @@ public UserDto registerUser(UserDto userDto) {
...
+    if (userDto != null) {
+      if (userDto.getEnabled() && userDto.getVerified()) {
+        throw new UserException(UserExceptionType.USER_ALREADY_REGISTERED);
+      }
+    }
 
+    String token = jwtEmailVerificationUtils.generateJwtEmailVerificationToken(email);
+
+    String confirmationLink = jwtEmailVerificationUtils.getConfirmationLink() + token;
+    String subject = "Confirm your email for SkyNest";
```
petarstoimenovskihtec: This can be extract as constant in this file or in general constants class.

backend/src/main/resources/application.properties
```diff
@@ -15,3 +15,16 @@ spring.data.cassandra.cluster-name=Test Cluster
 spring.data.cassandra.local-datacenter=datacenter1
 spring.data.cassandra.keyspace-name=tokenkeyspace
 spring.data.cassandra.schema-action=CREATE_IF_NOT_EXISTS
+
+#   App Properties - Email Jwt Verification Token
+backend.app.jwtEmailVerificationSecret=emailVerificationSecretKey
+backend.app.jwtEmailVerificationExpirationMs=900000
+backend.app.confirmationLink=http://localhost:8080/users/confirm?token=
```
petarstoimenovskihtec, line 35: Keep in mind that when you deploy this on server, the ip address (here localhost) will be the different

petarstoimenovskihtec (approved):

PR #23, opened 2022-06-02
SJQ2T-55: Better Cassandra init and scripts for Docker

ivanfhtec:


petarstoimenovskihtec (approved):
I will approve it for now, but will take a look at some time later in the sprints. I believe that some scripts are unnecessary, but it is ok for now

PR #24, opened 2022-06-02
Validation messages error messages redirect to login on register

casualme2:


marija-antic (approved):
LGTM. Focus on extracting components, I would suggest starting from atoms - buttons, inputs, etc.

PR #27, opened 2022-06-03
SJQ2T-57: Implement getUsers functionality BE

DarkoVeljanoskiHtec:
I've implemented GetUsers functionality, with the according changes in the userController and userService as well as test for the userService method. It doesn't return the users as of now because of the restricted security, but that should be solved during the Authorization implementation.


backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -48,9 +49,11 @@ public ResponseEntity<String> resendUserEmail(@RequestParam String email) {
     return ResponseEntity.ok(response);
   }
 
-  @GetMapping
-  public String getUser() {
-    //  TODO
-    return "test";
+  @GetMapping()
```
markobarackovhtec: you dont need brackets here

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -19,15 +19,19 @@
...
+import java.util.stream.Collectors;
+
 @Service
 @AllArgsConstructor
 public class UserServiceImpl implements UserService {
 
   private static final String SUBJECT_FOR_EMAIL_CONFIRMATION = "Confirm your email for SkyNest";
   private UserRepository userRepository;
-  private RoleService roleService;
+  private RoleRepository roleRepository;
```
markobarackovhtec: Where is this used ?

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -48,9 +49,11 @@ public ResponseEntity<String> resendUserEmail(@RequestParam String email) {
...
+  public List<UserResponse> getUsers() {
+    List<UserDto> listOfUsers = userService.listAllUsers();
+    return listOfUsers.stream()
+        .map(e -> modelMapper.map(e, UserResponse.class))
```
markobarackovhtec, line 74: In userService you are mapping it in UserDto, and then you are mapping to UserResponse, why are you using two model objects instead of one ?

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
```diff
@@ -181,4 +182,18 @@ void confirmEmail_UserWithEmailNotFound() {
     Assertions.assertThrows(
         UsernameNotFoundException.class, () -> userService.confirmEmail(anyString()));
   }
+
+  @Test
+  void listAllUsers() {
```
markobarackovhtec, line 178: This test is not doing anything, first UserEntity needs to have some attributes set, and then you can check if the response is mapped to correct return value with correct attributes

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
DarkoVeljanoskiHtec: Of course, thank you.

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
DarkoVeljanoskiHtec: roleRepository is never used so far, I'll remove it.

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
DarkoVeljanoskiHtec: That's how it was done for the registerUser method so I went along, should I map it to UserResponse directly in the service?

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
DarkoVeljanoskiHtec: Yes, I think I get it now, I'll try to change it.

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
markobarackovhtec: Yeah, you can do that in another PR, and to fix that for all other endpoints in UserController and to exclude model mapper from this class. You can create a ticket for that to do after this one.

DarkoVeljanoskiHtec: Sure, we are due for refactoring anyway :)

markobarackovhtec (approved):

markobarackovhtec:
I expect that next task is going to be "Exclude model mapper from controller class" :)

PR #30, opened 2022-06-03
SJQ2T-58 and SJQ2T-59: Added some Maven and Spring profiles

ivanfhtec:
Maven profiles:
* local (default)
* docker (skips tests)

The two Spring profiles have the same names.

The Maven profiles each inject the corresponding Spring profile into <code>application.properties</code>. The <code>local</code> profile has an additional <code>application-local.properties</code>.


backend/src/main/resources/application-local.properties
```diff
@@ -0,0 +1,13 @@
+#   local values
```
markobarackovhtec: Maybe is better to use ${FIELD_NAME} for specific values that could be different for yours local env, instead of real values

ivanfhtec: That makes sense. So we would all add environment variables to the run configuration/env vars instead.

Although it might be easier to do the same thing we did with the <code>.env</code> file: .gitignore this file and have a <code>application-local.properties.example</code> with all the necessary variables defined as blank. Then everyone would have make their own <code>application-local.properties</code>, or copy someone else's.

ivanfhtec: That makes sense. So we would all add environment variables to the run configuration/env vars instead.

Although it might be easier to do the same thing we did with the <code>.env</code> file: .gitignore this file and have a <code>application-local.properties.example</code> with all the necessary variables defined as blank. Then everyone would have make their own <code>application-local.properties</code>, or copy someone else's.

markobarackovhtec: As you wish, i will leave that part to you to decide

petarstoimenovskihtec: Well I personally don't like **_application-local.properties.example_** cause I think is unnecessary. If you want to take advantage of **_.env_**, here is a hint: take a look at 
```
<dependency>
		<groupId>me.paulschwarz</groupId>
		<artifactId>spring-dotenv</artifactId>
</dependency>
```

ivanfhtec: I'll take a look.

backend/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM maven:3.8.5-openjdk-11
 WORKDIR /backend
 COPY . .
-RUN mvn clean install -Dmaven.test.skip=true
-CMD mvn spring-boot:run
\ No newline at end of file
+RUN mvn clean install -Pdocker
+CMD mvn spring-boot:run -Pdocker
```
petarstoimenovskihtec: I don't believe you need to add profile when running spring app

backend/pom.xml
```diff
@@ -16,6 +16,26 @@
...
+            <id>local</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <properties>
+                <spring.profiles.active>local</spring.profiles.active>
+            </properties>
+        </profile>
+        <profile>
+            <id>docker</id>
```
petarstoimenovskihtec: Since this can be reused and it is not tightly-coupled with docker you can rename it to lets say to no-test or something like that

backend/pom.xml
```diff
@@ -37,6 +57,12 @@
             <groupId>org.springframework.boot</groupId>
             <artifactId>spring-boot-starter-security</artifactId>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/com.auth0/java-jwt -->
```
petarstoimenovskihtec, line 60: You guys really need to sort this formatting thing... @MetoMladenovski @DarkoVeljanoskiHtec @DraganMitrasinovicHtec @MarkoRadunovicHtec

petarstoimenovskihtec (commented):
Few comments

backend/Dockerfile
ivanfhtec: Since I've set <code>local</code> as the default Maven profile, without <code>-Pdocker</code> Maven injects <code>local</code> as the default Spring profile into <code>application.properties</code>.

backend/pom.xml
ivanfhtec: I thought we should have a dedicated profile for Docker, though right now it just has no-tests. I'll rename it.

backend/src/main/resources/application-local.properties
ivanfhtec: I can't really get this library to work - looks like <code>application.properties</code> keeps pulling vars from <code>.env</code> in root no matter what.

I wanted to just rename <code>application-local.properties</code> to <code>.env</code> and pull values from there (when running locally).

petarstoimenovskihtec: ping me tmr, we will take a look

backend/Dockerfile
petarstoimenovskihtec: Well then don't set `local` as default one :)

backend/pom.xml
petarstoimenovskihtec: There is an `Actions on Save` setting in IntelliJ that will automatically do that for you 

Screenshot 2022-06-06 at 18 18 54

petarstoimenovskihtec (changes requested):
[application-local.properties.example] is not the way to resolve this.. so pls change it

ivanfhtec:
I've just managed to make it work, will push commit soon

backend/pom.xml
```diff
@@ -16,6 +16,26 @@
...
+            </properties>
+        </profile>
+        <profile>
+            <id>no-tests</id>
+            <activation>
+                <activeByDefault>true</activeByDefault>
+            </activation>
+            <properties>
+                <maven.test.skip>true</maven.test.skip>
+                <spring.profiles.active>no-tests</spring.profiles.active>
```
petarstoimenovskihtec: OK I see what is the problem. By default **_no-tests_** maven profile is active if nothing is specified, that means that by default you are loading **_no-tests_** spring profile, so that will search for **_application-no-tests.properties_** which you do not have and parent configuration file will be loaded (application.properties)

backend/src/main/resources/application.properties
```diff
@@ -1,16 +1,7 @@
...
-TOKEN_DB_HOST=localhost
-TOKEN_DB_PORT=9042
-TOKEN_DB_USERNAME=cassandra
-TOKEN_DB_PASSWORD=cassandra
-TOKEN_DB_CLUSTER=Test Cluster
-TOKEN_DB_LOCAL_DATACENTER=datacenter1
-TOKEN_DB_KEYSPACE=tokenkeyspace
+spring.profiles.active=@spring.profiles.active@
+#   local env vars
+.env.filename=.local.env
```
petarstoimenovskihtec: well this way you are forcing the properties to be placed in **.local.env** file. Remember when running this code in some other env you won't have that file since it is in gitignore and do not need to force the env to be set in that specific way. For example if you deploy on cloud such as AWS, most of the time you will not have access to it and you can not specify that content of the file, and env variables will need to be loaded on some other way.

ivanfhtec: Where should we read environment variables from?

Currently, the values from <code>.local.env</code> are only active if there are no other sources for the necessary environment variables. This only happens if the app is running locally.

If the app is running from a docker container, these values are overwritten by values from <code>spring.env</code>, and this file pulls (only some) variables from <code>.env</code> in root.

petarstoimenovskihtec: Well env variables can be read from different places such as file like you have, docker variables, you can pass them as input params when you start the app from terminal, by docker override inline params in terminal as well, already set in the machine where the app will run, or some other external place such as cloud, aws, external system etc etc, you get the point :) Currently the code is only accepting env vars to be read from that hardcoded **_.local.env_**, so you do not have the flexibility that I mentioned above :)

petarstoimenovskihtec (approved):

PR #31, opened 2022-06-03
SJQ2T-65: hotfix CORS 

ivanfhtec:
Allow FE's origin for every endpoint.

Allow unauthenticated access to swagger.


DraganMitrasinovicHtec (approved):
Legit!

PR #32, opened 2022-06-06
SJQ2T-66: Swagger documentation

MarkoRadunovicHtec:
Swagger documentation edited trough annotations, Names edited, Example request and return values added for every endpoint

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
+++ b/backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
@@ -169,9 +169,7 @@ public class UserController {
               @Content(
                   mediaType = "application/json",
                   schema = @Schema(implementation = String.class),
-                  examples = {
-                    @ExampleObject(value = "The email failed to send from the Email server")
-                  })
+                  examples = {@ExampleObject(value = "Failed to send email")})
             })
       })
   @PostMapping("/resend-email")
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -18,11 +26,83 @@
...
+                  schema = @Schema(implementation = ErrorMessage.class),
+                  examples = {
+                    @ExampleObject(
+                        value =
+                            "{\"messages\":[\"This email is already in use\"],"
+                                + " \"status\": \"400\","
+                                + " \"timestamp\": \"2022-06-03 16:18:12\"}")
+                  })
+            })
+      })
```
petarstoimenovskihtec, line 95: You don't expect error 500 here? :P :D  Applies to every endpoint

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -18,11 +26,83 @@
...
+              schema = @Schema(implementation = UserRegisterRequest.class),
+              mediaType = "application/json",
+              examples = {
+                @ExampleObject(
+                    value =
+                        "{\"email\": \"username@gmail.com\","
+                            + " \"password\": \"paSword1\","
+                            + "  \"name\": \"Name\","
+                            + "  \"surname\": \"Surname\","
+                            + "  \"phoneNumber\": \"38166575757\","
```
petarstoimenovskihtec, line 108: Maybe add the phone format here if you have any ?

MarkoRadunovicHtec: When schema is opened in Request body, there is shown a pattern for phone number, which is only number. We didn't specify it any further

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
MarkoRadunovicHtec: Definitely :D Also I saw in example that I missed status code.. Fixed it

petarstoimenovskihtec: maybe for consistency use the same format as the other error 500 desc

petarstoimenovskihtec (approved):

PR #33, opened 2022-06-06
SJQ2T-55: Clean scripts now also remove backend image to force rebuild

ivanfhtec:


docker scripts/unix/clean-all.sh
```diff
@@ -2,6 +2,8 @@ echo "Removing all containers..."
 docker ps --all --quiet | xargs docker rm
 echo "Pruning everything..."
 docker system prune --volumes --force
+echo "Removing backend image..."
+docker image rm --force skynest/skynest-api
```
petarstoimenovskihtec, line 6: What if between stopping and restarting there are no new changes? Why forcing to rebuild ?

ivanfhtec: The <code>stop</code> script stops everything without forcing a rebuild.

The backend image is cached by default so even when the code changes, rebuilds don't happen. The rest of the team pulled the new code, but used the old image and thought we didn't fix anything.

petarstoimenovskihtec: anytime when you pull a new code and want to start the app via docker, then you need to rebuild the new image by using `docker-compose build `. I am just thinking that most of the team will not use this as intended but anyway it is good that you provide different options 💪

petarstoimenovskihtec (approved):

PR #37, opened 2022-06-06
SJQ2T-63 Delete User

MetoMladenovski:
Implemented delete user by Id.
Required by QA to delete automatically generated users, that are used for testing.

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
+++ b/backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
@@ -13,6 +13,7 @@ import org.springframework.http.ResponseEntity;
 import org.springframework.web.bind.annotation.*;
 
 import javax.validation.Valid;
+import java.util.UUID;
 import java.util.List;
 import java.util.stream.Collectors;
 
--- a/backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
+++ b/backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
@@ -15,6 +15,7 @@ import org.springframework.web.bind.annotation.*;
 import javax.validation.Valid;
 import java.util.UUID;
 import java.util.List;
+import java.util.UUID;
 import java.util.stream.Collectors;
 
 @RestController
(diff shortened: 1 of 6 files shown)


backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -74,4 +74,12 @@ public List<UserResponse> getUsers() {
         .map(e -> modelMapper.map(e, UserResponse.class))
         .collect(Collectors.toList());
   }
+
+  @DeleteMapping("/delete")
+  public ResponseEntity<String> deleteUser(@RequestParam String uuid) {
```
petarstoimenovskihtec: If this is UUID already, than change the type to it, instead of mapping on service layer

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -74,4 +74,12 @@ public List<UserResponse> getUsers() {
...
+    userService.deleteUser(uuid);
```
petarstoimenovskihtec, line 82: What happens if I try to delete a user that doesn't exist? Will it throw an error, or give OK status with successful message that user is deleted (when actually it doesn't exist in first place) or something else? How will you check that?

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
MetoMladenovski: I had some issues doing it this way, but now it's working just fine. I changed it

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
MetoMladenovski: I guess I forgot to handle the exception.. :( Will do it now.

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
markobarackovhtec: Maybe is good that uuid is actually path param instead of query param

petarstoimenovskihtec (approved):

PR #39, opened 2022-06-07
SJQ2T-18: Authorisation, JWT utils

ivanfhtec:
includes some refactoring


backend/src/main/java/com/htecgroup/skynest/SkyNestApplication.java
```diff
@@ -27,12 +27,12 @@ public void addCorsMappings(CorsRegistry registry) {
             .addMapping("/**")
             .allowedOrigins("http://localhost:3000")
             .allowedHeaders(
-                SecurityConstants.HEADER_STRING,
+                JwtUtils.HEADER_STRING,
```
petarstoimenovskihtec: Since you are refactoring this :D **_HEADER_STRING_** doesn't tell me anything, rename it to something more descriptive maybe **_AUTH_HEADER_**

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -264,14 +269,17 @@ public ResponseEntity<String> confirmPasswordReset(
   }
 
   @Operation(summary = "Get User with that id")
+  @PreAuthorize("hasAuthority(T(com.htecgroup.skynest.model.entity.RoleEntity).ROLE_WORKER)")
```
petarstoimenovskihtec, line 269: What if you need to extract this endpoint to be acceptable with additional role? Check @RolesAllowed

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -143,7 +148,7 @@ public ResponseEntity<UserResponse> registerUser(
                   })
             })
       })
-  @GetMapping("/confirm")
+  @GetMapping(CONFIRM_EMAIL_URL)
```
petarstoimenovskihtec, line 148: **Just a NOTE:** 
Endpoints naming needs work, one of the main things when API is created is naming of the API (meaning when you see the endpoint along with the type you know 99% what should be done). Also restructuring of this class is needed but this suggestions lets keep for some other iteration (user and auth actions should not be in the same class)

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -264,15 +269,18 @@ public ResponseEntity<String> confirmPasswordReset(
   }
 
   @Operation(summary = "Get User with that id")
+  @PreAuthorize("hasAuthority(T(com.htecgroup.skynest.model.entity.RoleEntity).ROLE_WORKER)")
   @GetMapping
   public List<UserResponse> getUsers() {
+    log.info("User {} accessed getusers", SecurityContextHolder.getContext().getAuthentication());
```
petarstoimenovskihtec: Well if you are invoking SecurityContext just to print it, it is better not to do it. If you want tho, then Spring automatically can inject the Principal / Authentication as input param in the controller method.
My advice if you want to print which user triggers which actions for every log statement in the project, then you need to do it with Spring security filter and MD5 .

You know the drill, technical dept task in backlog

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
ivanfhtec: We're only using authorities right now, so users don't have roles yet. Not sure how to do that.

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
ivanfhtec: I'll remove it, it was for debugging anyway.

backend/src/main/java/com/htecgroup/skynest/exception/UserExceptionHandler.java
```diff
@@ -1,29 +1,55 @@
...
+import java.io.IOException;
 import java.util.List;
+import java.util.Map;
 import java.util.stream.Collectors;
 
 @ControllerAdvice
 @Slf4j
 public class UserExceptionHandler {
 
+  public static ResponseEntity<ErrorMessage> responseFor(UserException ex) {
```
petarstoimenovskihtec: If you use this methods in some other classes, remove then from here.. `ContollerAdvice` has just one purpose and that is to handle all unhandled or purposely thrown exceptions and return generic response.

On the other hand if you want some method from ControllerAdvice to be executed, then throw that custom exception from that class that you handle

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -57,9 +59,13 @@ public Authentication attemptAuthentication(
     if (!passwordEncoder.matches(credentials.getPassword(), userDetails.getPassword())) {
       throw new UserException(UserExceptionType.PASSWORDS_DOES_NOT_MATCH);
     }
+
+    if (!userService.isActive(credentials.getEmail()))
```
petarstoimenovskihtec: What should be first - checking if the user is active or checking if the password matches?

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -69,16 +75,20 @@ protected void successfulAuthentication(
...
+    List<String> authorities =
+        user.getAuthorities().stream()
+            .map(GrantedAuthority::getAuthority)
+            .collect(Collectors.toList());
+    String token =
+        JwtUtils.generate(
+            user.getUsername(), JwtUtils.ACCESS_TOKEN_EXPIRATION_MS, "roles", authorities);
 
-    response.addHeader(SecurityConstants.HEADER_STRING, SecurityConstants.TOKEN_PREFIX + token);
+    response.addHeader(JwtUtils.HEADER_STRING, JwtUtils.TOKEN_PREFIX + token);
```
petarstoimenovskihtec: For string concatenation use String.format

backend/src/main/java/com/htecgroup/skynest/security/CustomUserDetailsService.java
```diff
@@ -20,6 +21,9 @@ public class CustomUserDetailsService implements UserDetailsService {
   @Override
   public UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
     UserDto userDto = userService.findUserByEmail(username);
-    return new User(userDto.getEmail(), userDto.getEncryptedPassword(), new ArrayList<>());
+    ArrayList<SimpleGrantedAuthority> authorities = new ArrayList<>();
+    authorities.add(new SimpleGrantedAuthority(userDto.getRole().getName()));
```
petarstoimenovskihtec: Can you rewrite this with java 8 ?

backend/src/main/java/com/htecgroup/skynest/exception/UserExceptionHandler.java
ivanfhtec: I wanted to catch and process our custom exceptions from the auth filter. Since the filter is not a service, the handler won't catch it.

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
```diff
@@ -21,36 +25,44 @@ public class WebSecurity extends WebSecurityConfigurerAdapter {
...
+        .antMatchers(HttpMethod.POST, UrlUtil.POST_URLS_WITHOUT_AUTH)
         .permitAll()
-        .antMatchers(HttpMethod.GET, SecurityConstants.CONFIRM_EMAIL_URL)
+        .antMatchers(HttpMethod.GET, UrlUtil.GET_URLS_WITHOUT_AUTH)
         .permitAll()
-        .antMatchers(HttpMethod.POST, SecurityConstants.PASSWORD_RESET_REQUEST_URL)
+        .antMatchers(HttpMethod.PUT, UrlUtil.PUT_URLS_WITHOUT_AUTH)
         .permitAll()
-        .antMatchers(HttpMethod.PUT, SecurityConstants.PASSWORD_RESET_CONFIRM_URL)
+        .antMatchers(HttpMethod.DELETE, UrlUtil.DELETE_URLS_WITHOUT_AUTH)
```
petarstoimenovskihtec: Is this empty for now?

backend/src/main/java/com/htecgroup/skynest/util/UrlUtil.java
```diff
@@ -0,0 +1,44 @@
...
+  public static final String LOG_IN_URL = "/login";
+  public static final String SIGN_UP_URL = "/register";
+  public static final String CONFIRM_EMAIL_URL = "/confirm";
+  public static final String RESEND_EMAIL_URL = "/resend-email";
+  public static final String PASSWORD_RESET_REQUEST_URL = "/password-reset/request";
+  public static final String PASSWORD_RESET_CONFIRM_URL = "/password-reset/confirm";
+
+  public static final String[] ANY_URLS_WITHOUT_AUTH = {SWAGGER_URL, SWAGGER_URL_ALT};
+
+  public static final String[] GET_URLS_WITHOUT_AUTH = {
```
petarstoimenovskihtec, line 23: Are you combining urls to use as one in spring security ? Or there is some other logic behind it?

backend/src/main/java/com/htecgroup/skynest/util/UrlUtil.java
```diff
@@ -0,0 +1,44 @@
...
+  };
+  public static final String[] POST_URLS_WITHOUT_AUTH = {
+    USERS_CONTROLLER_URL + SIGN_UP_URL, USERS_CONTROLLER_URL + PASSWORD_RESET_REQUEST_URL
+  };
+  public static final String[] PUT_URLS_WITHOUT_AUTH = {
+    USERS_CONTROLLER_URL + PASSWORD_RESET_CONFIRM_URL
+  };
+  public static final String[] DELETE_URLS_WITHOUT_AUTH = {};
+
+  public static final String[] URLS_WITHOUT_AUTH =
```
petarstoimenovskihtec, line 33: Pls pls find more clean solution... 


hint: java 8 Stream

backend/src/main/resources/application.properties
```diff
@@ -28,13 +27,15 @@ spring.data.cassandra.cluster-name=${TOKEN_DB_CLUSTER}
 spring.data.cassandra.local-datacenter=${TOKEN_DB_LOCAL_DATACENTER}
 spring.data.cassandra.keyspace-name=${TOKEN_DB_KEYSPACE}
 spring.data.cassandra.schema-action=CREATE_IF_NOT_EXISTS
-
+#   JWT
+jwt.access-expiration-ms=1800000
+jwt.secret=jfgi95jgnfi4
+jwt.prefix=Bearer
```
petarstoimenovskihtec: You have this in a constant right?

petarstoimenovskihtec (changes requested):
Few remarks

backend/src/main/java/com/htecgroup/skynest/exception/UserExceptionHandler.java
petarstoimenovskihtec: make it a component if that is the problem.. but since it is part of spring I don't see why wouldn't it catch it

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
ivanfhtec: It is, we'll add more to the list when needed.

backend/src/main/java/com/htecgroup/skynest/util/UrlUtil.java
ivanfhtec: yeah, pretty much

backend/src/main/resources/application.properties
ivanfhtec: It's loaded into a static field by Spring. Should it be a const?

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
ivanfhtec: It's refactored now.

backend/src/main/java/com/htecgroup/skynest/exception/UserExceptionHandler.java
ivanfhtec: I've tried adding Component or Service annotations to the filter, but it doesn't work.

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -18,16 +18,20 @@
...
-import java.util.UUID;
 import java.util.List;
 import java.util.UUID;
 import java.util.stream.Collectors;
 
+import static com.htecgroup.skynest.util.UrlUtil.*;
+
 @RestController
-@RequestMapping("/users")
+@EnableGlobalMethodSecurity(prePostEnabled = true, securedEnabled = true, jsr250Enabled = true)
```
petarstoimenovskihtec: The controller should not be annotated with spring security, for that purpose use new class for configuration

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
petarstoimenovskihtec: From the name that you use you already have roles :) so try with annotation that I sent.. here is a link with explanation https://stackoverflow.com/questions/43961625/rolesallowed-vs-preauthorize-vs-secured. From that link you will see what needs to be removed in the new configuration class

backend/src/main/java/com/htecgroup/skynest/exception/UserExceptionHandler.java
petarstoimenovskihtec: You can extract that in separate class then, but not like this now.. Maybe it will be good if creating od ErrorMessage is in util class in a method and you just reference to the same logic from both places

ivanfhtec: I've done some refactoring, but it's slightly different to what you suggested. It should be acceptable.

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -35,31 +36,34 @@ public class CustomAuthenticationFilter extends UsernamePasswordAuthenticationFi
...
-    log.info("The user: {} is trying to authenticate.", credentials.getEmail());
-    if (!credentials.getEmail().equals(userDetails.getUsername())) {
-      throw new UserException(UserExceptionType.INVALID_EMAIL_OR_PASSWORD);
-    }
+    log.info("The user: {} is trying to authenticate.", userDetails.getUsername());
+    if (!userService.isActive(userDetails.getUsername()))
+      throw new UserException(UserExceptionType.USER_NOT_ACTIVE);
 
-    if (!passwordEncoder.matches(credentials.getPassword(), userDetails.getPassword())) {
+    try {
```
petarstoimenovskihtec: There is no need of second try, you can merge everything into one try with multiple catch clauses

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -69,16 +73,20 @@ protected void successfulAuthentication(
...
       throws IOException, ServletException {
-    String userName = ((User) authentication.getPrincipal()).getUsername();
-    log.info("{} is successfully logged in.", userName);
 
-    Date expiration = new Date(System.currentTimeMillis() + SecurityConstants.EXPIRATION_TIME);
-    Algorithm algorithm = Algorithm.HMAC256(SecurityConstants.TOKEN_SECRET);
+    User user = (User) authentication.getPrincipal();
 
-    String token = JwtTokens.generateToken(userName, expiration, algorithm);
+    List<String> authorities =
```
petarstoimenovskihtec, line 77: This creation od authorities list is a good candidate to be in a separate method / Class.

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthorizationFilter.java
```diff
@@ -0,0 +1,67 @@
...
+      if (!authorizationHeader.startsWith(JwtUtils.TOKEN_PREFIX)) {
+        filterChain.doFilter(request, response);
+        return;
+      }
+
+      String token = authorizationHeader.replace(JwtUtils.TOKEN_PREFIX, "");
+      SecurityContextHolder.getContext().setAuthentication(JwtUtils.getFrom(token));
+
+      filterChain.doFilter(request, response);
+    } catch (UserException ex) {
```
petarstoimenovskihtec, line 51: just add a todo for this to be cleaned up in some next iteration

backend/src/main/java/com/htecgroup/skynest/security/CustomUserDetailsService.java
```diff
@@ -20,6 +21,10 @@ public class CustomUserDetailsService implements UserDetailsService {
   @Override
   public UserDetails loadUserByUsername(String username) throws UsernameNotFoundException {
     UserDto userDto = userService.findUserByEmail(username);
-    return new User(userDto.getEmail(), userDto.getEncryptedPassword(), new ArrayList<>());
+
+    List<SimpleGrantedAuthority> authorities =
+        List.of(new SimpleGrantedAuthority(userDto.getRole().getName()));
```
petarstoimenovskihtec: Since this list has only one item, usually best practice is to use Collections.singletonList

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
petarstoimenovskihtec: Then add this part when you needed, do not leave code that currently doesn't do anything. Pls remove it. Same applies to other permitAll matcher

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
```diff
@@ -21,36 +25,44 @@ public class WebSecurity extends WebSecurityConfigurerAdapter {
...
         .and()
         .addFilter(getAuthenticationFilter())
+        .addFilterBefore(
+            new CustomAuthorizationFilter(), UsernamePasswordAuthenticationFilter.class)
         .sessionManagement()
         .sessionCreationPolicy(SessionCreationPolicy.STATELESS);
   }
 
   public CustomAuthenticationFilter getAuthenticationFilter() throws Exception {
     final CustomAuthenticationFilter filter =
```
petarstoimenovskihtec, line 57: Since this filter already extends other filter that is part of Spring aka a Spring Bean, you can annotate this class as a spring component and inject it here.. The logic can be moved inside the class

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
ivanfhtec: We should probably refactor that and the JWT stuff at the same time

backend/src/main/resources/application.properties
petarstoimenovskihtec: well if the value is the same per env , then const is the way to do it

petarstoimenovskihtec (changes requested):
Few comments again :) getting there 👍

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
ivanfhtec: I've corrected it.

backend/src/main/java/com/htecgroup/skynest/security/CustomUserDetailsService.java
ivanfhtec: I've corrected it.

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
petarstoimenovskihtec: ok just mark it with todo, so when someone else is watching the code, to have the sense what needs to be done

ivanfhtec: Leave a // TODO in the code? Or do you mean as a task.

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
ivanfhtec: Working on it

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
petarstoimenovskihtec: yeah, leaving //todo in the code is completely fine and useful for future reference

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
```diff
@@ -11,6 +14,7 @@
 import org.springframework.security.config.annotation.web.configuration.WebSecurityConfigurerAdapter;
 import org.springframework.security.config.http.SessionCreationPolicy;
 import org.springframework.security.crypto.bcrypt.BCryptPasswordEncoder;
+import org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter;
 
 @EnableWebSecurity
 @Configuration
```
petarstoimenovskihtec, line 21: This class is a good candidate to move the configuration from UserController

ivanfhtec: I've moved it, but RolesAllowed always returns 403, and I made sure to test this jsr250Enabled=true.

petarstoimenovskihtec (approved):
All things that are left, will be done in some other PR

PR #40, opened 2022-06-07
SJQ2T-60: Fixed and added some tests

MetoMladenovski:
Fixed a couple of unfunctional tests and added a few new ones.


backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
```diff
@@ -90,8 +89,8 @@ void registerUser() {
     UserDto expectedUserDto = new ModelMapper().map(expectedUserEntity, UserDto.class);
 
     when(userRepository.existsByEmail(anyString())).thenReturn(false);
-    when(roleRepository.findByName(anyString()))
-        .thenReturn(Optional.of(new RoleEntity(UUID.randomUUID(), RoleEntity.ROLE_WORKER)));
+    when(roleService.findByName(anyString()))
+        .thenReturn(modelMapper.map(roleWorkerEntity, RoleDto.class));
```
petarstoimenovskihtec: if you are using modelMapper here just to create the response object, you can always return object that is mocked. such as mock(RoleDto.class)

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
```diff
@@ -114,17 +113,6 @@ void registerUser_AlreadyExists() {
     Assertions.assertThrows(UserException.class, () -> userService.registerUser(newUserDto));
```
petarstoimenovskihtec: Maybe also add a verification that corresponding mock services are called and that the return error message is as expected

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
```diff
@@ -203,4 +191,29 @@ void resetPassword() {
 
     Assertions.assertEquals(expectedResponse, userService.resetPassword(anyString(), anyString()));
   }
+
+  @Test
+  void deleteUser_UserWithIdNotFound() {
+    when(userRepository.existsById(any())).thenReturn(false);
+
+    Assertions.assertThrows(UserException.class, () -> userService.deleteUser(null));
```
petarstoimenovskihtec: some mock value is always better then **_null_**, because null can have some different meaning

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
```diff
@@ -203,4 +191,29 @@ void resetPassword() {
...
+  }
+
+  @Test
+  void enableUser() {
+    UserDto notEnabledUserDto = modelMapper.map(enabledWorkerEntity, UserDto.class);
```
petarstoimenovskihtec, line 194: same here.. you don;t need modelMapper to create the UserDto object, you can use the corresponding constructor.
P.S. What is the default Boolean value when object has some property from that type?

MetoMladenovski: There is an object UserEntity created in the setUp method. So I thought I can just map that userEntity in a userDto with the same values and not create a new one. And that UserEntity has the enabled and verified set to true, that is why I am setting them to false.

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
MetoMladenovski: Because the method deleteUser accepts UUID, i can't mock it. I made it generate a UUID that should never be in our database.

petarstoimenovskihtec: This is unit test, so no database is included here :) You are mocking here what will happen if **_existsById_** returns false so any value can be used. If you need UUID , you can always create one with `UUID.randomUUID()`

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
petarstoimenovskihtec: if that UserEntity is created just for mapping to userDto, maybe you can consider create userDto instead?

petarstoimenovskihtec:
There are lots of classes that were already merged with development branch, those files should not be part of this PR, pls take a look how to fix that

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
MetoMladenovski: The UserEntity is used in previous test cases, I just thought it would be nice to reuse it.

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
MetoMladenovski: You are completely right

PR #43, opened 2022-06-08
SJQ2T-93: Exclude ModelMapper from controller class

DarkoVeljanoskiHtec:
This is not done, the tests needs to be changed because of changes in the main classes, i just wanted to get a feedback is this any good, or should we disscard this completely.


backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -114,9 +110,8 @@ public class UserController {
   public ResponseEntity<UserResponse> registerUser(
       @Valid @RequestBody UserRegisterRequest userRegisterRequest) {
 
-    UserDto userDto = userService.registerUser(modelMapper.map(userRegisterRequest, UserDto.class));
-
-    return new ResponseEntity<>(modelMapper.map(userDto, UserResponse.class), HttpStatus.OK);
+    ResponseEntity<UserResponse> responseEntity = new ResponseEntity<>(userService.registerUser(userRegisterRequest), HttpStatus.OK);
+    return responseEntity;
```
ivanfhtec, line 122: Lines 113 and 114 could be one line

ivanfhtec (commented):
These changes are in response to [this](https://github.com/JDP2022SkyNest/SkyNest/pull/27#discussion_r889904019)?

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
DarkoVeljanoskiHtec: Yes that's why i was doing this task, yes i guess it could be one line, but this way it's more consistent with the type of code we are writing imo.

petarstoimenovskihtec (approved):

petarstoimenovskihtec:
We will refactor the app in some other iteration or when you have more time

PR #45, opened 2022-06-08
SJQ2T-95 Testing UserServiceImpl

MetoMladenovski:
Fixed and added some new tests for UserServiceImpl class.


backend/src/test/java/com/htecgroup/skynest/utils/UserDtoUtil.java
```diff
@@ -0,0 +1,31 @@
...
+        currentDateTime,
+        null,
+        email,
+        password,
+        encryptedPassoword,
+        name,
+        surname,
+        address,
+        phoneNumber,
+        true,
```
petarstoimenovskihtec: should this values be false instead?

MetoMladenovski: Yes they should

petarstoimenovskihtec (approved):

PR #48, opened 2022-06-10
SJQ2T-134: Cassandra Entities and Repository for InvalidJWTs

ivanfhtec:
Use the InvalidJwtService to invalidate tokens and check if tokens are invalid.

Important: multiple invocations of <code>invalidate()</code> for the same token will not refresh the time to live in Cassandra.


backend/src/main/java/com/htecgroup/skynest/repository/custom/impl/CassandraSaveWithTtlImpl.java
```diff
@@ -0,0 +1,25 @@
...
+
+@AllArgsConstructor
+public class CassandraSaveWithTtlImpl<T> implements CassandraSaveWithTtl<T> {
+
+  private CassandraOperations operations;
+
+  @Override
+  public <S extends T> S save(S entity, int timeToLiveInSeconds) {
+
+    Assert.notNull(entity, "Entity must not be null");
```
petarstoimenovskihtec: Instead of this, you can take a look at annotations such as _@NotNull_ and _@Min_ and apply that validation here

backend/src/main/java/com/htecgroup/skynest/repository/custom/impl/CassandraSaveWithTtlImpl.java
```diff
@@ -0,0 +1,25 @@
...
+import org.springframework.data.cassandra.core.InsertOptions;
+import org.springframework.util.Assert;
+
+@AllArgsConstructor
+public class CassandraSaveWithTtlImpl<T> implements CassandraSaveWithTtl<T> {
+
+  private CassandraOperations operations;
+
+  @Override
+  public <S extends T> S save(S entity, int timeToLiveInSeconds) {
```
petarstoimenovskihtec, line 14: What if I want this object to be persisted without timeToLive, meaning for good or until I explicitly request to delete it? Can we provide that option as well since you already used this generic approach ?

backend/src/main/java/com/htecgroup/skynest/model/entity/InvalidJwtEntity.java
```diff
@@ -0,0 +1,17 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.NoArgsConstructor;
+import lombok.Setter;
+import org.springframework.data.cassandra.core.mapping.PrimaryKey;
+import org.springframework.data.cassandra.core.mapping.Table;
+
+@Table
+@Getter
+@Setter
```
petarstoimenovskihtec: Why do you need the setter?

backend/src/main/java/com/htecgroup/skynest/repository/custom/impl/CassandraSaveWithTtlImpl.java
ivanfhtec: InvalidJwtRepository also extends CassandraRepository, which already has a regular <code>save()</code> that persists indefinitely.

backend/src/main/java/com/htecgroup/skynest/repository/custom/impl/CassandraSaveWithTtlImpl.java
ivanfhtec: I borrowed this code from the <code>save()</code> method from <code>SimpleCassandraRepository.java</code>. I'll switch to use annotations since they'll look cleaner.

backend/src/main/java/com/htecgroup/skynest/model/entity/InvalidJwtEntity.java
ivanfhtec: I thought I would, but I don't. I'll remove it.

backend/src/main/java/com/htecgroup/skynest/repository/custom/impl/CassandraSaveWithTtlImpl.java
ivanfhtec: I've removed validation here as it's done in the service anyway.

petarstoimenovskihtec (approved):

PR #49, opened 2022-06-10
SJQ2T-94: BE - Refactor env vars, JWT and email utils

DraganMitrasinovicHtec:
Moved sensitive info to .env and spring.env.
Reformatted our Jwt classes, now checking for the claim when validating  tokens.
Reformatted the Email sending service, now using .html files as templates


backend/src/main/java/com/htecgroup/skynest/model/email/Email.java
```diff
@@ -1,18 +1,14 @@
...
 import lombok.Getter;
 
 @Getter
+@AllArgsConstructor
 public class Email {
   String to;
   String subject;
-  String emailBody;
+  String template;
+  String link;
```
petarstoimenovskihtec: What happens if you need to sent a mail where no link option is needed? That option will be null ? Also what happens if there is email where you need to provide some user data or any other data that currently you don't support in this Email class?
Try to find more elegant and extendable solution for this kind of situations

DraganMitrasinovicHtec: Yeah, this won't be too reusable in the future, as it is specific for the 2 emails we are currently sending. Will think of a better way.

DraganMitrasinovicHtec: @petarstoimenovskihtec Would it be sufficient to have an array of String args, which could be filled with any info required by the email template you are using? So when you make a new email template which, for example, needs the users name, you can add it and use it in the template as args[2] ( or whichever index it is at ).

DraganMitrasinovicHtec: @petarstoimenovskihtec Would it be sufficient to have an array of String args, which could be filled with any info required by the email template you are using? So when you make a new email template which, for example, needs the users name, you can add it and use it in the template as args[2] ( or whichever index it is at ).

DraganMitrasinovicHtec:
This should be better, added a Map to the email class so you can add any arguments you would like to use in the template there.

backend/src/main/java/com/htecgroup/skynest/model/email/Email.java
petarstoimenovskihtec: there is a saying in IT world.. when you are in doubt use Map :D

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -284,10 +277,8 @@ public ResponseEntity<String> confirmPasswordReset(
   @PreAuthorize("hasAuthority(T(com.htecgroup.skynest.model.entity.RoleEntity).ROLE_WORKER)")
```
petarstoimenovskihtec, line 281: I know this is not part of your PR, but why is this method annotated with Role-Worker ? isn't it already by default ?

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -67,35 +73,49 @@ public void sendVerificationEmail(String emailAddress) {
...
+            arguments,
+            true));
   }
 
   @Override
-  public String resetPassword(String token, String password) {
-    emailUtils.validateJwtToken(token);
-    String email = emailUtils.getEmailFromJwtEmailToken(token);
+  public String resetPassword(UserPasswordResetRequest userPasswordResetRequest) {
+    JwtUtils.validateEmailToken(
```
petarstoimenovskihtec: Maybe you can provide static method that will call this method under the hood, for example JwtUtils.validateTokenForPasswordReset

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -116,13 +136,13 @@ public void deleteUser(UUID uuid) {
   }
 
   @Override
-  public UserDto getUser(UUID uuid) {
+  public UserResponse getUser(UUID uuid) {
```
petarstoimenovskihtec, line 73: Wasn't this already refactored?

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -136,17 +156,17 @@ public UserDto findUserByEmail(String email) {
...
-        .map(e -> modelMapper.map(e, UserDto.class))
+        .map(e -> modelMapper.map(e, UserResponse.class))
         .collect(Collectors.toList());
   }
 
   @Override
   public String confirmEmail(String token) {
-    emailUtils.validateJwtToken(token);
-    String email = emailUtils.getEmailFromJwtEmailToken(token);
+    JwtUtils.validateEmailToken(token, JwtUtils.EMAIL_VERIFICATION_PURPOSE);
```
petarstoimenovskihtec: Since you are refactoring this and this both methods are doing pretty much the same thing, can you merge them into one method that will validate the token and return the subject or whatever is more suitable?

backend/src/main/java/com/htecgroup/skynest/util/EmailUtils.java
```diff
@@ -1,214 +1,17 @@
...
-        + "    <tr>\n"
-        + "      <td height=\"30\"><br></td>\n"
-        + "    </tr>\n"
-        + "  </tbody></table><div class=\"yj6qo\"></div><div class=\"adL\">\n"
-        + "\n"
-        + "</div></div>";
-  }
+  public static final String EMAIL_VERIFICATION_SUBJECT = "Password reset for SkyNest";
+  public static final String PASSWORD_RESET_SUBJECT = "Password reset for SkyNest";
+  public static final String EMAIL_VERIFICATION_TEMPLATE = "emails/verification";
```
petarstoimenovskihtec: this class is a good candidate to be extracted in enum, meaning you only need to provide the enum key, and it will give you the template and the subject. That way you will reduce the margin for error (for example use EMAIL_VERIFICATION_SUBJECT and PASSWORD_RESET_TEMPLATE as a couple in some method), and always use the correct value pairs

backend/src/main/java/com/htecgroup/skynest/util/JwtUtils.java
```diff
@@ -28,8 +31,13 @@ public class JwtUtils {
   public static final String TOKEN_PREFIX = "Bearer ";
 
   public static long ACCESS_TOKEN_EXPIRATION_MS;
+  public static long EMAIL_TOKEN_EXPIRATION_MS;
   private static Algorithm ALGORITHM;
 
+  private static final String EMAIL_TOKEN_CLAIM = "Email Token";
+  public static final String EMAIL_VERIFICATION_PURPOSE = "verification";
```
petarstoimenovskihtec: Does this needs to be in the this util class, or it can be exported in a constants class? Or maybe in the new enum class if it makes sense?

backend/src/main/java/com/htecgroup/skynest/util/UrlUtil.java
```diff
@@ -2,9 +2,13 @@
...
 import lombok.NoArgsConstructor;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.stereotype.Component;
+import org.springframework.web.servlet.support.ServletUriComponentsBuilder;
 
 import java.util.stream.Stream;
 
+@Component
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
 public class UrlUtil {
```
petarstoimenovskihtec: good practice is if the constructor is private to mark the class as final, meaning it is closed for extensions.

backend/src/main/java/com/htecgroup/skynest/util/UrlUtil.java
```diff
@@ -39,4 +45,23 @@ public class UrlUtil {
...
           .toArray(String[]::new);
+
+  public static String getEmailVerificationLink(String token) {
+    return String.format(
+        "%s%s%s%s%s",
+        getBaseUrlBackend(), USERS_CONTROLLER_URL, CONFIRM_EMAIL_URL, "?token=", token);
+  }
+
+  public static String getPasswordResetLink(String token) {
+    return String.format("%s%s%s%s", BASE_URL_FRONTEND, "/confirm-password", "?token=", token);
```
petarstoimenovskihtec: Since you already have constants for other urls, do the same here.. and maybe define them as path, cause it is just the part of the url , right?

backend/src/main/resources/templates/emails/password-reset.html
```diff
@@ -0,0 +1,94 @@
+<!DOCTYPE html>
```
petarstoimenovskihtec, line 1: I like that this is extracted in new html file 👍 Maybe we can use thymeleaf in future (do not change it now)

petarstoimenovskihtec (changes requested):
Few changes

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
DraganMitrasinovicHtec: It was, rebased too early by accident. Will check with the team if that is necessary

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
DraganMitrasinovicHtec: Agreed, so the logic doesn't leak into UserService.

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
DraganMitrasinovicHtec: It was, rebased by accident so it's showing in this PR too.

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
DraganMitrasinovicHtec: Of course, the second method is 1 line anyways, makes more sense if that's just the return value.

backend/src/main/java/com/htecgroup/skynest/util/EmailUtils.java
DraganMitrasinovicHtec: Yeah, it sounds convenient to couple those for easier usage

backend/src/main/java/com/htecgroup/skynest/util/JwtUtils.java
DraganMitrasinovicHtec: I don't think it belongs in the new enum, as I will be using the enum for sending emails, not verification. Does it make sense to make a whole another class for constants instead of leaving it here? @petarstoimenovskihtec

backend/src/main/java/com/htecgroup/skynest/util/UrlUtil.java
DraganMitrasinovicHtec: Will do

DraganMitrasinovicHtec: Yeah, everything in the class is just the part of the url, will look to rename everything.

backend/src/main/resources/templates/emails/password-reset.html
DraganMitrasinovicHtec: Thanks :) , that was the main motivation behind the refactor

backend/src/main/java/com/htecgroup/skynest/util/JwtUtils.java
petarstoimenovskihtec: If it is used in multiple different places, then constant class is good candidate

backend/src/main/java/com/htecgroup/skynest/service/UserService.java
```diff
@@ -1,31 +1,34 @@
...
+import com.htecgroup.skynest.model.response.UserResponse;
+import com.htecgroup.skynest.model.request.UserPasswordResetRequest;
 
 import java.util.List;
 import java.util.UUID;
 
 public interface UserService {
 
-  UserDto registerUser(UserDto userDto);
+  UserResponse registerUser(UserRegisterRequest userRegisterRequest);
```
petarstoimenovskihtec, line 12: Since this interface and class impl with it is becoming more and more bigger, maybe it is good time to separate things between user manipulation and auth logic. Having that said, all methods that are connected by any means with auth such as (confirmEmail, verifyUser, email sending methods, password change etc) should go in new class.

backend/src/main/java/com/htecgroup/skynest/service/impl/EmailServiceImpl.java
```diff
@@ -19,17 +21,22 @@
...
   @Async
   public void send(Email email) {
+    Context context = new Context();
+    context.setVariable("args", email.getArgs());
+
     try {
       MimeMessage mimeMessage = javaMailSender.createMimeMessage();
       MimeMessageHelper helper = new MimeMessageHelper(mimeMessage, "utf-8");
-      helper.setText(email.getEmailBody(), email.getHtml());
+      helper.setText(templateEngine.process(email.getEmailType().getTemplate(), context), true);
```
petarstoimenovskihtec: everytime when you have same chained method calls, it is good to extract that in separate variable, in this case: email.getEmailType()

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -67,35 +72,35 @@ public void sendVerificationEmail(String emailAddress) {
       throw new UserException(UserExceptionType.USER_ALREADY_REGISTERED);
     }
 
-    String token = emailUtils.generateJwtEmailToken(emailAddress);
+    String token = JwtUtils.generateEmailToken(emailAddress, JwtUtils.EMAIL_VERIFICATION_PURPOSE);
+
+    Map<String, String> arguments =
+        Map.of("emailAddress", emailAddress, "link", UrlUtil.getEmailVerificationLink(token));
```
petarstoimenovskihtec: why do you need the emailAddress property in the map, when you propagate it as separate field in Email constructor? Btw you can always extract Email creation in separate method for logic separation and better readability.. even new EmailHelper class is perfectly fine if it makes sense to you

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -67,35 +72,35 @@ public void sendVerificationEmail(String emailAddress) {
...
     }
 
-    String token = emailUtils.generateJwtEmailToken(email);
+    String token = JwtUtils.generateEmailToken(emailAddress, JwtUtils.PASSWORD_RESET_PURPOSE);
 
-    String confirmationLink = emailUtils.getPasswordResetLink() + token;
-    String emailBody = emailUtils.buildPasswordResetEmail(email, confirmationLink);
-    Email emailToSend = new Email(email, SUBJECT_FOR_PASSWORD_RESET, emailBody, true);
-    emailService.send(emailToSend);
+    Map<String, String> arguments =
```
petarstoimenovskihtec: Same applies here

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -67,35 +72,35 @@ public void sendVerificationEmail(String emailAddress) {
...
+
+    emailService.send(new Email(emailAddress, EmailType.PASSWORD_RESET, arguments, true));
   }
 
   @Override
-  public String resetPassword(String token, String password) {
-    emailUtils.validateJwtToken(token);
-    String email = emailUtils.getEmailFromJwtEmailToken(token);
+  public String resetPassword(UserPasswordResetRequest userPasswordResetRequest) {
+    String email = JwtUtils.validatePasswordResetToken(userPasswordResetRequest.getToken());
```
petarstoimenovskihtec: when validate method is used, usually it is expected a boolean type to be returned. In your case since you return the email value, maybe rename the method to something more descriptive.. **_getValidatedResetPasswordTokenContent_** or something like it

petarstoimenovskihtec: Same can and should be applied to every other method from that kind

backend/src/main/java/com/htecgroup/skynest/util/JwtUtils.java
```diff
@@ -58,13 +66,49 @@ public static UsernamePasswordAuthenticationToken getFrom(String token) {
...
+    return validateEmailToken(token, JwtUtils.PASSWORD_RESET_PURPOSE);
+  }
+
+  public static String validateEmailVerificationToken(String token) {
+    return validateEmailToken(token, JwtUtils.EMAIL_VERIFICATION_PURPOSE);
+  }
+
   @Value("${jwt.access-expiration-ms}")
-  public void setNameStatic(long expirationMs) {
+  private void setAccessTokenExpirationMs(long expirationMs) {
```
petarstoimenovskihtec, line 109: you can always set that value in field instead of method, meaning 
```
@Value("${jwt.access-expiration-ms}")
private final Long ACCESS_TOKEN_EXPIRATION_MS;
```

I think that was the correct syntax

backend/src/main/resources/application.properties
```diff
@@ -11,6 +11,11 @@ TOKEN_DB_PASSWORD=cassandra
 TOKEN_DB_CLUSTER=Test Cluster
 TOKEN_DB_LOCAL_DATACENTER=datacenter1
 TOKEN_DB_KEYSPACE=tokenkeyspace
+JWT_SECRET=jfgi95jgnfi4
+ACCESS_EXPIRATION_MS=1800000
+EMAIL_EXPIRATION_MS=1800000
+SPRING_MAIL_USERNAME=no.reply.skynest@gmail.com
+SPRING_MAIL_PASSWORD=ftzldfejzhtqjsok
```
petarstoimenovskihtec: This values I believe should be set in .env file, by this I mean: JWT_SECRET and SPRING_MAIL_PASSWORD. Others can be left hard coded as they previously were.

The point of using this .env variables is to extract the sensitive information in separate place so they will not be attached with the code, but there reference will be used instead.

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
```diff
@@ -82,7 +82,7 @@ void setUp() {
     newUserRequest.setAddress("Address");
   }
 
-  @Test
+  /*@Test
   void registerUser() {
```
petarstoimenovskihtec, line 86: if it is not used and won't be used in future, you can delete it. Same applies for every commented test

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
```diff
@@ -201,16 +199,16 @@ void listAllUsers() {
...
+            .map(e -> modelMapper.map(e, UserResponse.class))
             .collect(Collectors.toList());
 
-    List<UserDto> returnedDTOs = userService.listAllUsers();
+    List<UserResponse> returnedResponse = userService.listAllUsers();
 
-    Assertions.assertEquals(expectedDTOs.size(), returnedDTOs.size());
-    Assertions.assertEquals(expectedDTOs.get(0), returnedDTOs.get(0));
-    Assertions.assertEquals(expectedDTOs.get(0).getEmail(), returnedDTOs.get(0).getEmail());
+    Assertions.assertEquals(expectedResponse.size(), returnedResponse.size());
```
petarstoimenovskihtec, line 209: There is a good code structure the @MetoMladenovski did in his PR, maybe you need to change that according to his changes when he merges the PR.

backend/src/main/java/com/htecgroup/skynest/service/UserService.java
DraganMitrasinovicHtec: Agreed, we were thinking of doing that anyways. Better to just do it now

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
DraganMitrasinovicHtec: Forgot to change it, I intended to send the Users name in the arguments, sounds weird referring to them by their email.

DraganMitrasinovicHtec: @petarstoimenovskihtec  Do you mean something like createVerificationEmail() in some sort of helper or util class?

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
DraganMitrasinovicHtec: Yeah, current name doesn't indicate that it's returning the email

backend/src/main/java/com/htecgroup/skynest/util/JwtUtils.java
DraganMitrasinovicHtec: I tried that first but it doesn't work with static fields. So I won't be able to use the variable it in static methods

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
DraganMitrasinovicHtec: Yeah, there will be lots of test refactoring. Is it fine if I do that in a separate PR?

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
petarstoimenovskihtec: yup, something like that

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
petarstoimenovskihtec: yeah sure

DraganMitrasinovicHtec:
Branch is causing me trouble as it is very out of date, will make a new one with the same changes, possibly split it up into smaller ones.

PR #50, opened 2022-06-10
SJQ2T-127: BE - Check role for user info

MarkoRadunovicHtec:
Checks for roles to access getUser added, service class added for accessing current user info. Need to add tests


backend/src/main/java/com/htecgroup/skynest/service/impl/CurrentUserServiceImpl.java
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  @Override
+  public String getEmail() {
+    return getAuthentication().getPrincipal().toString();
+  }
+
+  @Override
+  public String getRole() {
+    return new ArrayList<>(getAuthentication().getAuthorities()).get(0).getAuthority();
```
petarstoimenovskihtec: What if the user has more than one Role? Maybe you can change this method to return List instead

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -117,6 +119,14 @@ public void deleteUser(UUID uuid) {
 
   @Override
   public UserDto getUser(UUID uuid) {
```
petarstoimenovskihtec: What is the purpose of this method?

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
```diff
@@ -73,6 +75,23 @@ void setUp() {
             roleWorkerEntity,
             null);
 
+    otherWorkerEntity =
+        new UserEntity(
+            UUID.randomUUID(),
```
MetoMladenovski: I think you can use enabledWorkerEntity, no need to create another one. 
It will need to change anyway, because I made some changes to the tests structure, but still.

backend/src/main/java/com/htecgroup/skynest/service/impl/CurrentUserServiceImpl.java
```diff
@@ -0,0 +1,45 @@
...
+import java.util.UUID;
+
+@Service
+@AllArgsConstructor
+public class CurrentUserServiceImpl implements CurrentUserService {
+
+  private UserRepository userRepository;
+
+  @Override
+  public Authentication getAuthentication() {
```
petarstoimenovskihtec: Consider the option to return the logged User object, (maybe that will trigger some changes in that model as well, like extending spring security User object) and instead of 3 methods getAuthentication, getEmail and getRole, you will have just one and that will be getLoggedUser()

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
MarkoRadunovicHtec: I used it when I needed to check that other user entity is returned when admin calls get user. I cant check the constraint with the same entity. I'll take a look anyway.

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
MarkoRadunovicHtec: It is used for accessing and later changing user personal data. If he is worker he only can access his data, but admin can access every user data.

backend/src/main/java/com/htecgroup/skynest/service/impl/CurrentUserServiceImpl.java
MarkoRadunovicHtec: Sure, will do it

backend/src/main/java/com/htecgroup/skynest/service/impl/CurrentUserServiceImpl.java
MarkoRadunovicHtec: I'll take a look how it can be done.

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
petarstoimenovskihtec: so having that said:  what will this line do when I invoke it for my user 
`**_if (currentRole.equals(RoleEntity.ROLE_WORKER) && !(currentUuid.equals(uuid)))_**, `

I am a user with worker roles, that means part 1 of the if clause is ok, and my id is the same one as the one from the session, meaning part 2 `!(currentUuid.equals(uuid)` is false? that will mean I won't be able to see my data right?

Same applies if I am with admin role :D

MarkoRadunovicHtec: Yes, that's right. If user has worker role but wants to see other user personal data he won't be able to do so.

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
```diff
@@ -148,7 +179,21 @@ void getUser_IdNotFound() {
...
-  void getUser() {
+  void getUser_Admin() {
+    when(userRepository.findById(any())).thenReturn(Optional.of(otherWorkerEntity));
+    when(currentUserService.getRole()).thenReturn("role_admin");
+    when(currentUserService.getId()).thenReturn(enabledWorkerEntity.getId());
+
+    UserDto expectedUserDto = userService.getUser(otherWorkerEntity.getId());
+
+    Assertions.assertEquals(
+        new ModelMapper().map(otherWorkerEntity, UserDto.class), expectedUserDto);
```
petarstoimenovskihtec: with this new ModelMapper() you are tightly coupling the test to use the same dependency as the actual implementation. You can always assert based on every attribute that each object contains. For reference you can see the changes that are applied or Metodi will apply in his PR for tests

MarkoRadunovicHtec: I'll take a look :)

backend/src/test/java/com/htecgroup/skynest/service/impl/UserServiceImplTest.java
petarstoimenovskihtec: check Metodis PR

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
petarstoimenovskihtec: so this means according to the logic that if I am Admin , an Exception would be thrown? and that was not the wanted case, right?

MarkoRadunovicHtec: No, if admin access his or other user info exception would not be thrown. Only if worker try to access other user info exception is thrown.

backend/src/main/java/com/htecgroup/skynest/security/CustomAuthenticationToken.java
```diff
@@ -0,0 +1,31 @@
...
+   *     by this authentication object.
+   */
+  public CustomAuthenticationToken(
+      LoggedUserDto loggedUserDto, Collection<? extends GrantedAuthority> authorities) {
+    super(loggedUserDto.getUsername(), loggedUserDto.getPassword(), authorities);
+    this.loggedUserDto = loggedUserDto;
+  }
+
+  @Override
+  public Object getPrincipal() {
```
petarstoimenovskihtec: You can use LoggedUserDto instead of Object

MarkoRadunovicHtec: Fixed it :)

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -125,6 +129,15 @@ public void deleteUser(UUID uuid) {
 
   @Override
   public UserResponse getUser(UUID uuid) {
+
+    LoggedUserDto loggedUserDto = currentUserService.getLoggedUser();
+    String loggedUserRole = new ArrayList<>(loggedUserDto.getAuthorities()).get(0).toString();
```
petarstoimenovskihtec: You can create a method in LoggedUserDto that will accept userRole and the logic will be to check if user has that role.. Something like **_hasRole(String role)_** or whatever type you need as input param. That way that logic will be reusable, this code will be there and you can call that method with `loggedUserDto.hasRole(RoleEntity.ROLE_WORKER.name()) `

MarkoRadunovicHtec: Yes, that would be nice. Will do something like that

petarstoimenovskihtec (approved):

PR #55, opened 2022-06-14
SJQ2T-150: Change regex for password 

MetoMladenovski:
Changed regex for password to be able to accept special characters.


DraganMitrasinovicHtec (approved):
Seems good

PR #56, opened 2022-06-14
SJQ2T-143 and SJQ2T-147: CORS for deployment and env vars for URLs

ivanfhtec:
* CORS allows server's origin
* localhost can always go through CORS, for testing purposes
* added url-related environment variables


ivanfhtec:
Force pushed rebase to development.

ivanfhtec:
Force pushed rebase to development.

backend/src/main/java/com/htecgroup/skynest/SkyNestApplication.java
```diff
@@ -24,11 +25,15 @@ public WebMvcConfigurer corsConfigurer() {
...
       public void addCorsMappings(CorsRegistry registry) {
+        String publicIp = UrlUtil.SKYNEST_HOST;
+        String frontendPort = String.valueOf(UrlUtil.FRONTEND_PORT);
         registry
             .addMapping("/**")
             .allowedMethods("*")
-            .allowedOrigins("http://localhost:3000")
-            .allowedOriginPatterns("http://localhost:3000/*")
+            .allowedOrigins(
+                "http://localhost:3000", String.format("http://%s:%s", publicIp, frontendPort))
```
petarstoimenovskihtec: so you already have that values now in separate variables right? Why don't you use them instead of the hard coded ones.

backend/src/main/resources/application-local.properties
```diff
@@ -1,12 +1,4 @@
+#   environment variables
+.env.prefix=
```
petarstoimenovskihtec, line 2: This works as previous?

backend/src/main/java/com/htecgroup/skynest/util/UrlUtil.java
```diff
@@ -39,4 +42,23 @@ public class UrlUtil {
               new String[] {USERS_CONTROLLER_URL + LOG_IN_URL})
           .flatMap(Stream::of)
           .toArray(String[]::new);
+
+  public static String SKYNEST_HOST;
+  public static int BACKEND_PORT;
+  public static int FRONTEND_PORT;
+
+  @Value("${skynest.host}")
```
petarstoimenovskihtec: you can use ConfigurationProperties annotation and see how can you extract this in separate property class

backend/src/main/java/com/htecgroup/skynest/SkyNestApplication.java
ivanfhtec: It's hard-coded so that so that FE could send requests from their local builds. These requests would have <code>Origin: http://localhost:3000</code>.

backend/src/main/resources/application-local.properties
ivanfhtec: Yes, all the env vars are visible without the <code>env.</code> prefix, which makes the rest of <code>application-local.properties</code> identical to <code>application.properties</code>.

The only difference is the JDBC string. SSL between Docker containers is a pain to set up, so it's turned off. But if the app is running locally there's no reason to turn it off.

backend/src/main/java/com/htecgroup/skynest/SkyNestApplication.java
petarstoimenovskihtec: Again this way you are hardcoding into the app, not based on environment where it is loaded.
The point is this to be restricted based on specific env.
In this case if you add skynest.host=http://localhost in the application-local.properties, and if FE is starting the app locally with corresponding profile - local, as they should, it will work right? 

On the other hand, when you deploy this to the VM it will work with value set in ${SKYNEST_HOST}, so no changes will be done in the code, right?

petarstoimenovskihtec: from the second look, if you delete the hardcoded "http://localhost:3000" here, and only apply the change in application-local.properties your generic code will work out of the box

ivanfhtec: Yeah, it's probably simpler if FE just tests their local build with our local build. The deployed versions should only work with each other anyway.

ivanfhtec: I've removed the hard-coded bit. The should just change their <code>.env</code>.

backend/src/main/java/com/htecgroup/skynest/util/UrlUtil.java
ivanfhtec: I've moved these properties to a class annotated with ConfigurationProperties.

petarstoimenovskihtec (approved):

PR #58, opened 2022-06-15
SJQ2T-154: BE Controllers Refactor

DarkoVeljanoskiHtec:
made the changes so that the public endpoints starts with /public


ivanfhtec (approved):
Looks good.

PR #59, opened 2022-06-15
Admin panel

casualme2:
Promote and Demote Buttons are just there visually.
For now (since roles aren't implemented yet) admin can be accessed by everyone. 
As soon as we get that roles update, i'll change it.

Still more features need to be added as soon as we get that BE Update
I just want a feedback so far.


frontend/src/Components/AdminPanel/AdminPanel.css
```diff
@@ -0,0 +1,20 @@
+body {
```
marija-antic: Add changes to body somewhere higher so it can be visible. Maybe index.scss

frontend/src/Components/Navbar/Navbar.css
```diff
@@ -0,0 +1,101 @@
...
+   background-color: var(--mate-black);
+   width: 300px;
+   height: 100vh;
+   display: flex;
+   position: fixed;
+   top: 0;
+   left: -100%;
+   transition: 850ms;
+   z-index: 4;
+ }
```
marija-antic: Space between classes.

marija-antic (approved):
Looks good

PR #60, opened 2022-06-15
SJQ2T-76: Invalidate access tokens

MetoMladenovski:
Created new controller Logout and implemented endpoint logout. Now invalid tokens get stored in database until they expire and are filtered in CustomAuthorizationFilter.

This is the main idea behind logout. I will need the https://github.com/JDP2022SkyNest/SkyNest/pull/50 PR to be merged first so I can fix some stuff (like not using autowired).


backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthorizationFilter.java
```diff
@@ -43,8 +51,14 @@ protected void doFilterInternal(
         filterChain.doFilter(request, response);
         return;
       }
-
       String token = authorizationHeader.replace(JwtUtils.TOKEN_PREFIX, "");
+
+      if (invalidJwtService.isInvalid(token)) {
+        filterChain.doFilter(request, response);
+        log.info("Token is invalid, because it's present in invalidated tokens database.");
```
casual033: consider adding info about token, otherwise is not very useful

backend/src/main/java/com/htecgroup/skynest/controller/LogoutController.java
```diff
@@ -0,0 +1,52 @@
...
+                  examples = {@ExampleObject(value = "Logout successful")})
+            }),
+      })
+  @PostMapping
+  public ResponseEntity<String> logout(HttpServletRequest request) {
+    String authorizationHeader = request.getHeader(JwtUtils.AUTH_HEADER);
+    String token = authorizationHeader.replace(JwtUtils.TOKEN_PREFIX, "");
+    invalidJwtService.invalidate(token);
+    String response = "User logged out successfully";
+    log.info(response);
```
casual033: consider adding user id to message

casual033 (approved):

backend/src/main/java/com/htecgroup/skynest/controller/LogoutController.java
```diff
@@ -0,0 +1,50 @@
...
+            description = "Logout successful",
+            content = {
+              @Content(
+                  mediaType = "application/json",
+                  schema = @Schema(implementation = String.class),
+                  examples = {@ExampleObject(value = "Logout successful")})
+            }),
+      })
+  @PostMapping
+  public ResponseEntity<String> logout(HttpServletRequest request) {
```
ivanfhtec: Couldn't you just use the RequestHeader annotation?

MetoMladenovski: I guess we can

backend/src/main/java/com/htecgroup/skynest/controller/LogoutController.java
```diff
@@ -0,0 +1,49 @@
...
+                  schema = @Schema(implementation = String.class),
+                  examples = {@ExampleObject(value = "Logout successful")})
+            }),
+      })
+  @PostMapping
+  public ResponseEntity<String> logout(
+      @RequestHeader(name = "Authorization") String authorizationHeader) {
+    String token = authorizationHeader.replace(JwtUtils.TOKEN_PREFIX, "");
+    invalidJwtService.invalidate(token);
+    return ResponseEntity.ok("User logged out successfully");
```
casual033: I don't think the message is needed or useful here. If we want to show the message to user, it shouldn't be something hardcoded on backend.

casual033 (approved):

backend/src/main/java/com/htecgroup/skynest/controller/LogoutController.java
petarstoimenovskihtec: you can set as logger info statement here and maybe return true value as indicator that action went well, and false otherwise if needed

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthorizationFilter.java
```diff
@@ -20,6 +22,12 @@
 @Log4j2
 public class CustomAuthorizationFilter extends OncePerRequestFilter {
 
+  // I'm using @Autowired because using constructor DI doesn't work.
+  // I think it has something to do with the way CustomAuthorizationFilter is defined as a bean in
+  // the application class, as it calls the NoArgsConstructor and the dependencies are not loaded.
+  // This way the service is autowired and it works.
+  @Autowired private InvalidJwtService invalidJwtService;
```
petarstoimenovskihtec: if CustomAuthorizationFilter is defined as bean, you can inject InvalidJwtService in the class where CustomAuthorizationFilter is defined and pass that reference in the constructor. That way you can use the lombok @AllArgsConstructor and avoid Autowire annotation here.

petarstoimenovskihtec: with other words:
 @Bean
  public CustomAuthorizationFilter customAuthorizationFilter(InvalidJwtService invalidJwtService) {
    return new CustomAuthorizationFilter(invalidJwtService);
  }
and use standard constructor injection here with Lombok.

On the other hand you can create  new configuration class (annotated with @Configuration) and move there all beans defined in SkyNestApplication.class. That way you can separate the bean creation logic from application start

petarstoimenovskihtec (changes requested):
small but crucial change :)

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthorizationFilter.java
MetoMladenovski: Yes, it was already done in another PR, so I'am just gonna use it now.

petarstoimenovskihtec (approved):

PR #61, opened 2022-06-15
SJQ2T-157: BE - fix serialization issues

ivanfhtec:
* added no args constructors where necessary
* log ip address on login


MetoMladenovski (approved):
Should be fine now,

DarkoVeljanoskiHtec:
Looks like it should work

PR #67, opened 2022-06-16
SJQ2T-152: BE - Refresh Token Endpoint

DarkoVeljanoskiHtec:
made the refresh token endpoint as well as implementation, when calling the endpoint new access and refresh token are created.
the expiration time for the refresh token is about 20 days.


backend/src/main/java/com/htecgroup/skynest/service/impl/RefreshTokenServiceImpl.java
```diff
@@ -0,0 +1,59 @@
...
+@AllArgsConstructor
+@Log4j2
+public class RefreshTokenServiceImpl implements RefreshTokenService {
+
+  private UserService userService;
+  private ModelMapper modelMapper;
+
+  @Override
+  public void refreshToken(HttpServletRequest request, HttpServletResponse response) {
+    String header = request.getHeader("refresh-token");
```
petarstoimenovskihtec: You method should not be dependent on the request and response. The logic of the method is to refresh the token, meaning that you need the logged user as input param, and custom object that will contain the new attributes (new auth token, and refresh token ) . Then in the constructor you will apply that result in the header or body or how ever you need it.
Btw logged user you can see how to get it with the new change that @MarkoRadunovicHtec  did

petarstoimenovskihtec (changes requested):
Pls decouple service logic from request and response http objects

backend/src/main/java/com/htecgroup/skynest/service/impl/RefreshTokenServiceImpl.java
DarkoVeljanoskiHtec: I finally did it, it should be fine now.

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -255,4 +258,45 @@ public ResponseEntity<String> deleteUser(@PathVariable UUID uuid) {
...
+                  examples = {
+                    @ExampleObject(value = "Failed to send new access and refresh tokens")
+                  })
+            })
+      })
+  @GetMapping("/token/refresh")
+  public void refreshToken(
+      @RequestHeader("refresh-token") String refresh_token, HttpServletResponse response) {
+    String token = refreshTokenService.refreshToken(refresh_token);
+    response.addHeader("Authorization", "Bearer " + token);
```
ivanfhtec: Shouldn't this use constants from JwtUtils? Like in CustomAuthenticationFilter.

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -255,4 +258,45 @@ public ResponseEntity<String> deleteUser(@PathVariable UUID uuid) {
...
+    response.addHeader("refresh_token", refresh_token);
```
ivanfhtec: "refresh_token" should probably also be a constant

backend/src/main/java/com/htecgroup/skynest/service/impl/RefreshTokenServiceImpl.java
```diff
@@ -0,0 +1,50 @@
...
+@AllArgsConstructor
+@Log4j2
+public class RefreshTokenServiceImpl implements RefreshTokenService {
+
+  private CurrentUserService currentUserService;
+  private ModelMapper modelMapper;
+
+  @Override
+  public String refreshToken(String refresh_token) {
+    if (refresh_token != null && refresh_token.startsWith("Bearer ")) {
```
ivanfhtec: JwtUtils.TOKEN_PREFIX instead of  "Bearer "

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -89,8 +89,13 @@ protected void successfulAuthentication(
     String token =
         JwtUtils.generate(
             user.getUsername(), JwtUtils.ACCESS_TOKEN_EXPIRATION_MS, "roles", authorities);
+    String refresh_token =
+        JwtUtils.generate(
```
petarstoimenovskihtec: There should be new method for this when you take the latest development version

backend/src/main/java/com/htecgroup/skynest/service/impl/RefreshTokenServiceImpl.java
```diff
@@ -0,0 +1,50 @@
...
+      try {
+        LoggedUserDto loggedUserDto = currentUserService.getLoggedUser();
+
+        List<String> authorities =
```
petarstoimenovskihtec: Again there is a new method for this

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -86,7 +86,11 @@ protected void successfulAuthentication(
 
     String token = JwtUtils.generateAccessToken(jwtObject, authorities);
 
+    String refresh_token = JwtUtils.generateRefreshToken(jwtObject, authorities);
+
     response.addHeader(JwtUtils.AUTH_HEADER, String.format("%s%s", JwtUtils.TOKEN_PREFIX, token));
+    response.addHeader(
+        JwtUtils.REFRESH_TOKEN_HEADER, String.format("%s%s", JwtUtils.TOKEN_PREFIX, refresh_token));
```
ivanfhtec, line 90: Do we need add this header to CORS?

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -258,4 +262,46 @@ public ResponseEntity<String> deleteUser(@PathVariable UUID uuid) {
...
+                  mediaType = "application/json",
+                  schema = @Schema(implementation = String.class),
+                  examples = {
+                    @ExampleObject(value = "Failed to send new access and refresh tokens")
+                  })
+            })
+      })
+  @GetMapping("/token/refresh")
+  public void refreshToken(
+      @RequestHeader("refresh-token") String refresh_token, HttpServletResponse response) {
```
ivanfhtec: I'm not sure this header will go through CORS.

Also, how will this request pass through our authorization filter? The access token is already expired.

backend/src/main/java/com/htecgroup/skynest/service/impl/RefreshTokenServiceImpl.java
```diff
@@ -0,0 +1,43 @@
...
+@Service
+@AllArgsConstructor
+@Log4j2
+public class RefreshTokenServiceImpl implements RefreshTokenService {
+
+  private CurrentUserService currentUserService;
+
+  @Override
+  public String refreshToken(String refreshToken) {
+    if (refreshToken != null && refreshToken.startsWith(JwtUtils.TOKEN_PREFIX)) {
```
ivanfhtec, line 26: I don't think refreshToken can ever be null here.

Since you're using the RequestHeader annotation in the controller, if the request doesn't contain this header at all won't Spring throw an error without even getting here?

If the header is present, then refreshToken won't be null anyway.

backend/src/main/resources/application.properties
```diff
@@ -34,6 +34,7 @@ login-attempts.max-attempts-per-interval=5
 jwt.access-expiration-ms=${JWT_ACCESS_EXPIRATION}
 jwt.email-expiration-ms=${EMAIL_EXPIRATION}
 jwt.secret=${JWT_SECRET}
+jwt.refresh-expiration-ms=1800000000
```
ivanfhtec, line 37: this could probably be an env var

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
ivanfhtec: JwtUtils.REFRESH_TOKEN_HEADER instead of hard-coding

DarkoVeljanoskiHtec: This must be done while access token is still active, if not it can't pass you're right.

backend/src/main/resources/application.properties
DarkoVeljanoskiHtec: we'll add it in the next refactor

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
ivanfhtec: I guess you could just expect the refresh token in the authorization header here. Then it would just magically work with both CORS and the authorization filter. Not sure the mentors would like that though.

petarstoimenovskihtec: you are right , I don't like that hack at all :) it will work sure, but it is not the correct approach..

petarstoimenovskihtec: I will try to answer that in the PR it self

petarstoimenovskihtec:
The flow that you need to achieve:

1. User logins into the application (username and password)
2. BE returns any required information and
 a) Access JWT token with an expired time usually (15, 30 min).
 b) Refresh JWT token with an expired time greater than access one.

From now, the client app aka FE will use access token in the Authorization header for every request.

When BE returns 401, the FE app will try to use refresh token (using an specific endpoint - public accessible ) to get new credentials, without forcing the user to login again.

If there is no problem, new access token should be returned and user will be able to continue using the application. 
If BE returns a new 401 ->  FE should redirect to login page.

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
petarstoimenovskihtec: Here it is: https://github.com/JDP2022SkyNest/SkyNest/pull/67#issuecomment-1162275844

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
```diff
@@ -42,6 +42,8 @@ protected void configure(HttpSecurity http) throws Exception {
         .authorizeRequests()
         .antMatchers(String.format("%s/**", UrlUtil.PUBLIC_CONTROLLER_URL))
         .permitAll()
+        .antMatchers("/token/refresh")
+        .permitAll()
```
ivanfhtec, line 46: or you could change the authorization filter to use the RefreshToken header when servletPath is /token/refresh

this makes it seem like the client isn't authorized at all, when the client is just using a refresh token to authorize

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
DarkoVeljanoskiHtec: I've change it, I think it follows the flow we need to achieve now.

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthorizationFilter.java
```diff
@@ -43,7 +43,13 @@ protected void doFilterInternal(
         return;
       }
 
-      String authorizationHeader = request.getHeader(JwtUtils.AUTH_HEADER);
+      String authorizationHeader = "";
```
markobarackovhtec: you dont need quotes here
You can only say String authorizationHeader;

backend/src/main/java/com/htecgroup/skynest/service/impl/RefreshTokenServiceImpl.java
markobarackovhtec: It depends on required flag in RequestHeader annotation. if required is true then spring will throw exception if required is false then headed will be null

DarkoVeljanoskiHtec: I haven't specified is it required or not, not sure what the default value is, I think is required true.

markobarackovhtec: Still this check could stay here, no need to remove it

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
petarstoimenovskihtec: no no.. don't do that :) the main reason why refresh token is used is to create new access token, that is... don't change the flow :) btw I already explained here how should work: https://github.com/JDP2022SkyNest/SkyNest/pull/67#issuecomment-1162275844

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthorizationFilter.java
```diff
@@ -43,7 +43,13 @@ protected void doFilterInternal(
         return;
       }
 
-      String authorizationHeader = request.getHeader(JwtUtils.AUTH_HEADER);
+      String authorizationHeader;
+      if (request.getServletPath().startsWith(UrlUtil.REFRESH_TOKEN)) {
+        authorizationHeader = request.getHeader(JwtUtils.REFRESH_TOKEN_HEADER);
```
petarstoimenovskihtec: Why ?

backend/src/main/java/com/htecgroup/skynest/service/impl/RefreshTokenServiceImpl.java
```diff
@@ -0,0 +1,47 @@
...
+        String newAccessToken = JwtUtils.generateAccessToken(jwtObject, authorities);
+
+        return newAccessToken;
+      } catch (Exception exception) {
+        log.error(exception.getMessage(), exception);
+      }
+    } else {
+      throw new UserException(UserExceptionType.REFRESH_TOKEN_IS_MISSING);
+    }
+    return "";
```
petarstoimenovskihtec: Maybe it is better to return null or exception here, so the check on client side is easier

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthorizationFilter.java
DarkoVeljanoskiHtec: It was so the Authorization goes through Refresh token when the refresh endpoint is called, because the access token will be expired then, but according to your previous comment, I'm changing it now.

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
DarkoVeljanoskiHtec: The problem is if the access token is expired and the endpoint is public I can't take the current logged user info and if I don't have that, I can't make new access token.

petarstoimenovskihtec: but the refresh token needs to have the user id right? so you can extract that from there

backend/src/main/java/com/htecgroup/skynest/service/impl/RefreshTokenServiceImpl.java
```diff
@@ -0,0 +1,51 @@
...
+  private InvalidJwtServiceImpl invalidJwtService;
+
+  @Override
+  public String refreshToken(String refreshToken) {
+    if (refreshToken != null && refreshToken.startsWith(JwtUtils.TOKEN_PREFIX)) {
+      try {
+        String refresh_token = refreshToken.replace(JwtUtils.TOKEN_PREFIX, "");
+        String username = JwtUtils.getUsernameFromRefreshToken(refresh_token);
+        List<String> authorities =
+            JwtUtils.getAuthoritiesFromRefreshToken(refresh_token).stream()
```
petarstoimenovskihtec: Since authorities can be changed from the time when refresh token was issued, it is better to read them from UserDto if you can istead

backend/src/main/java/com/htecgroup/skynest/service/impl/RefreshTokenServiceImpl.java
```diff
@@ -0,0 +1,50 @@
...
+
+  @Override
+  public String refreshToken(String refreshToken) {
+    if (refreshToken != null && refreshToken.startsWith(JwtUtils.TOKEN_PREFIX)) {
+      try {
+        String refresh_token = refreshToken.replace(JwtUtils.TOKEN_PREFIX, "");
+        String username = JwtUtils.getUsernameFromRefreshToken(refresh_token);
+        UserDto userDto = userService.findUserByEmail(username);
+        List<String> authorities = new ArrayList<>();
+        authorities.add(userDto.getRole().getName());
```
petarstoimenovskihtec: use SingletonList instead

petarstoimenovskihtec (approved):
Fix and commit

PR #76, opened 2022-06-17
SJQ2T-161: Added UUID to JWT payload

MarkoRadunovicHtec:
Added UUID of logged user to JWT payload, getRoleNames method added to LoggedUserDto and corresponding usage of that method implemented into CustomAuthenticationFilter


backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -80,15 +79,13 @@ protected void successfulAuthentication(
...
 
-    List<String> authorities =
-        user.getAuthorities().stream()
-            .map(GrantedAuthority::getAuthority)
-            .collect(Collectors.toList());
+    Map<String, String> payload = Map.of("uuid", user.getUuid().toString());
     String token =
         JwtUtils.generate(
-            user.getUsername(), JwtUtils.ACCESS_TOKEN_EXPIRATION_MS, "roles", authorities);
+            user.getUsername(), payload, JwtUtils.ACCESS_TOKEN_EXPIRATION_MS, "roles", authorities);
```
petarstoimenovskihtec: I am just thinking at loud :) 

Is it possible to combine, username, and this uuid that is set in the map, into a custom object and send that object in the JwtUtils.generate method? Does that makes sense ?

backend/src/main/java/com/htecgroup/skynest/util/JwtUtils.java
```diff
@@ -38,9 +39,14 @@ public class JwtUtils {
   public static Algorithm ALGORITHM;
 
   public static String generate(
```
petarstoimenovskihtec: like in other cases, you can create a separate method for role claim. For example generateRoleClaim method that will call this generic method , that way "roles" as claim will be inside that new method and you will have control what is sent as claim instead of giving the option for free text

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
MarkoRadunovicHtec: I'll take a look

backend/src/main/java/com/htecgroup/skynest/util/JwtUtils.java
MarkoRadunovicHtec: Will try to do something like that

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -80,15 +79,14 @@ protected void successfulAuthentication(
...
+    List<String> authorities = user.getRoleNames();
 
-    List<String> authorities =
-        user.getAuthorities().stream()
-            .map(GrantedAuthority::getAuthority)
-            .collect(Collectors.toList());
-    String token =
-        JwtUtils.generate(
-            user.getUsername(), JwtUtils.ACCESS_TOKEN_EXPIRATION_MS, "roles", authorities);
+    JwtObject jwtObject = new JwtObject();
```
petarstoimenovskihtec: You can create a all args constructor and set those props there instead of setter here. having that said, setter is not needed

backend/src/main/java/com/htecgroup/skynest/model/jwtObject/JwtObject.java
```diff
@@ -0,0 +1,11 @@
+package com.htecgroup.skynest.model.jwtObject;
+
+import lombok.Data;
+
+import java.util.UUID;
+
+@Data
```
petarstoimenovskihtec: Instead of @Data, use just those that are needed

MarkoRadunovicHtec: Will fix it.

backend/src/test/java/com/htecgroup/skynest/service/impl/InvalidJwtServiceImplTest.java
```diff
@@ -39,6 +41,7 @@ void setUp() {
     token =
         JwtUtils.generate(
             "ivan.fajgelj@htecgroup.com",
+            Map.of("uuid", UUID.randomUUID().toString()),
```
petarstoimenovskihtec: You can change in the test to use this new custom object as well

petarstoimenovskihtec (approved):

PR #78, opened 2022-06-17
SJQ2T-167: BE - Change user role script

ivanfhtec:
* added stored procedure to change user role

* added convenience script to call said procedure


backend/src/main/resources/db/migration/V2.10__procedure_change_user_role.sql
```diff
@@ -0,0 +1,16 @@
+DELIMITER //
+CREATE PROCEDURE `change_user_role` (IN user_email CHAR(254), IN new_role CHAR(255))
```
petarstoimenovskihtec: Do you need a procedure for this? remember procedure exist in the database, and can be executed again? Do you need this? Maybe just apply the update scripts as regular sql script, it should work ok, since this is one time job

docker scripts/unix/change-user-role.sh
```diff
@@ -0,0 +1,11 @@
+echo "available roles:"
+echo -e "\trole_worker"
+echo -e "\trole_manager"
+echo -e "\trole_admin"
+read -p "new role: role_" role
+read -p "email: " username
+echo "--DB password"
+docker exec -it skynest-db mysql -u user -p --execute="USE skynest; CALL change_user_role('$username','role_$role')"
```
petarstoimenovskihtec: So you want to create a script that will be used to change the role dynamically on whatever user you like?

backend/src/main/resources/db/migration/V2.10__procedure_change_user_role.sql
ivanfhtec: Since we decided that we'll add/promote admins manually, this is mainly designed to promote existing users to admins. But we could also then reuse this code if we decide to add an endpoint for this, so I opted for a procedure.

Also, I wanted to display whether the operation succeeded or not in a reasonably clean way.

Should I just move the update query to the bash script for now?

docker scripts/unix/change-user-role.sh
ivanfhtec: Initially, I just wanted a script to promote existing users to admins. Then I realised that I should probably have one to demote admins as well. In the end, it turned into something completely generic.

backend/src/main/resources/db/migration/V2.10__procedure_change_user_role.sql
ivanfhtec: If the procedure stays, updateCount should probably be an OUT param instead of a local var.

petarstoimenovskihtec (approved):
Just extend the README file with info how to use this script

PR #79, opened 2022-06-17
SJQ2T-162: BE - Break down user service

DraganMitrasinovicHtec:
Refactored UserService into AuthService and UserService


backend/src/main/java/com/htecgroup/skynest/service/impl/AuthServiceImpl.java
```diff
@@ -0,0 +1,92 @@
...
+    Email email = EmailUtil.createVerificationEmail(userDto, token);
+
+    emailService.send(email);
+  }
+
+  @Override
+  public void sendPasswordResetEmail(String emailAddress) {
+    UserDto userDto = userService.findUserByEmail(emailAddress);
+
+    if (!userRepository.existsByEmail(emailAddress)) {
```
petarstoimenovskihtec: Basicly if userDto is not null it exists right?

DraganMitrasinovicHtec: Yeah, I checked it twice by accident, findUserByEmail already checks if the user exists, and throws an error if he doesn't.
Will remove the redundant check.

markobarackovhtec (approved):

markobarackovhtec:
Solve conflicts before merging

PR #80, opened 2022-06-17
SJQ2T-170: BE - Wrong Status Code

DarkoVeljanoskiHtec:
now it returns status code 401 when there isn't token provided and 403 when token is invalid.

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
+++ b/backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
@@ -9,6 +9,7 @@ import com.htecgroup.skynest.service.UserService;
 import com.htecgroup.skynest.util.UrlUtil;
 import lombok.RequiredArgsConstructor;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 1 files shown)


backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
```diff
@@ -45,6 +47,9 @@ protected void configure(HttpSecurity http) throws Exception {
         .and()
         .addFilter(getAuthenticationFilter())
         .addFilterBefore(getAuthorizationFilter(), UsernamePasswordAuthenticationFilter.class)
+        .exceptionHandling()
+        .authenticationEntryPoint(new HttpStatusEntryPoint(HttpStatus.UNAUTHORIZED))
```
ivanfhtec, line 51: elegant

DraganMitrasinovicHtec: exquisite

DraganMitrasinovicHtec (approved):
That's all that needs to be done, great

PR #81, opened 2022-06-20
SJQ2T-174 BE - Edit user fix

MarkoRadunovicHtec:
AuthorizeAccessToUserDetailsWith method added to UserService, added call of that method into UserController in getUser and editUser, added and modified some tests


DraganMitrasinovicHtec (approved):
Seems good, approving per Petar's request

PR #84, opened 2022-06-21
SJQ2T-176: BE - Admin edit fix

MarkoRadunovicHtec:
authorize method for admin trying to edit another admin info added to UserService, called in UserController on editUserUserService, called in UserController on editUser,


backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -204,7 +204,7 @@ public ResponseEntity<UserResponse> getUser(@PathVariable UUID uuid) {
   @PutMapping("/{uuid}")
   public ResponseEntity<UserResponse> editUser(
       @Valid @RequestBody UserEditRequest userEditRequest, @PathVariable UUID uuid) {
-    userService.authorizeAccessToUserDetailsWith(uuid);
+    userService.authorizeEditUserDetailsWith(uuid);
```
petarstoimenovskihtec: Discussed internally how this can be achieved with custom annotation. @DarkoVeljanoskiHtec pls sync with Marko and apply the same changes in your PR

backend/src/main/java/com/htecgroup/skynest/annotation/validator/EditValidator.java
```diff
@@ -0,0 +1,40 @@
...
+import lombok.RequiredArgsConstructor;
+import org.springframework.stereotype.Component;
+
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+import java.util.UUID;
+
+@Component
+@RequiredArgsConstructor
+public class EditValidator implements ConstraintValidator<CurrentUserCanEdit, UUID> {
```
petarstoimenovskihtec, line 17: what is the difference between this and ViewValidator? I believe you can use the same validator, and the same annotation, just override with Worker/Admin role in its usage definition

MarkoRadunovicHtec: In ViewValidator validation is that worker cannot see any other user details than his, in EditValidator there need to be checked that if admin is trying to edit or delete another admin it cant be done. So it needs to check if the accessed uuid from param(User with that Id) is admin than his info cannot be edited. But admin can view another admin info, just cant delete or edit other than his. So thats why there are two validator classes

MarkoRadunovicHtec: I forgot to mention that he can edit or delete another worker, but not another admin. I don't see the solution to use one validator class to include this separate checks. Because for different roles view can be accessed but not editing part. Admin can see every other user but can not edit another admin.

MarkoRadunovicHtec: Maybe to add access field into annotation class and then to pass Edit or View to it so it knows witch call is made and than in validator class to have switch or something like that to do different checks based on access field value.

petarstoimenovskihtec: I would keep everything separate.. Meaning separate validator for edit , for view and for delete..
For edit and delete shouldn't be dependent of user role, just the uuid, if I understood correctly.
For view is dependent of user role

backend/src/main/java/com/htecgroup/skynest/annotation/validator/DeleteValidator.java
```diff
@@ -0,0 +1,39 @@
...
+@Component
+@RequiredArgsConstructor
+public class DeleteValidator implements ConstraintValidator<CurrentUserCanDelete, UUID> {
+  private final CurrentUserService currentUserService;
+  private final UserService userService;
+  private CurrentUserCanDelete currentUserCanDelete;
+
+  @Override
+  public void initialize(CurrentUserCanDelete constraintAnnotation) {
+    currentUserCanDelete = constraintAnnotation;
```
petarstoimenovskihtec: Are you using this annotation reference anywhere?

backend/src/main/java/com/htecgroup/skynest/annotation/validator/EditValidator.java
```diff
@@ -0,0 +1,40 @@
...
+
+  private final CurrentUserService currentUserService;
+  private final UserService userService;
+  private CurrentUserCanEdit currentUserCanEdit;
+
+  @Override
+  public void initialize(CurrentUserCanEdit constraintAnnotation) {
+    currentUserCanEdit = constraintAnnotation;
```
petarstoimenovskihtec: same here

petarstoimenovskihtec (approved):
Some changes

PR #85, opened 2022-06-21
SJQ2T-179: BE - Admin delete fix

DarkoVeljanoskiHtec:
fixed using authorizeDeleteUserWith method in userService, now admin can't delete another admin.

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
+++ b/backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
@@ -208,7 +208,7 @@ public class UserServiceImpl implements UserService {
   }
 
   @Override
-  public void authorizeDeleteUserDetailsWith(UUID uuid) {
+  public void authorizeDeleteUserWith(UUID uuid) {
     try {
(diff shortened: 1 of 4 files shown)


backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -186,12 +186,33 @@ public UserDto verifyUser(UserDto userDto) {
     return userDto.withEnabled(true).withVerified(true);
   }
 
-  public void authorizeAccessToUserDetailsWith(UUID uuid) {
+  public void authorizeViewUserDetailsWith(UUID uuid) {
+    LoggedUserDto loggedUserDto = currentUserService.getLoggedUser();
```
petarstoimenovskihtec, line 190: Try to rewrite this with custom annotation instead. Applies for all new methods. Also try to create the methods reusable, this way you have lots of duplicate code

DarkoVeljanoskiHtec: I'll try that, thanks.

PR #88, opened 2022-06-23
SJQ2T-133: BE - Better errors

DraganMitrasinovicHtec:
Trying to substitute our error codes with correct ones. Not using enums anymore, instead extending the same base class and throwing static final exceptions.


backend/src/main/java/com/htecgroup/skynest/exception/EmailException.java
```diff
@@ -0,0 +1,15 @@
+package com.htecgroup.skynest.exception;
+
+import org.springframework.http.HttpStatus;
+
+public class EmailException extends CustomException {
+  private EmailException(String message, HttpStatus status) {
+    super(message, status);
+  }
+
+  public static final EmailException NON_EXISTING_EMAIL_ADDRESS =
```
petarstoimenovskihtec: Instead of static fields, it is always better to have appropriate Exception.. 
Let me give you example.
Here you can create NonExistingEmailException that will extend EmailException. You create just a default constructor that will call the super constructor with the attributes needed. The message can be created as a constant private field here , and status set directly.. 

This way you don't need to create static fields that will live in the app cycle and they will be created before used if used at all. With this direct appropriate exceptions, you create them as needed and it is easier for testing and more readable, meaning if you test EmailException for non existing email address, you need to provide the SendingEmailFailed field as well (and you don't need it). Applies for every other exception

DraganMitrasinovicHtec: Makes sense, so it uses lazy initialization. Was worried that would be too many classes, but I guess that doesn't really matter. Will change it everywhere

petarstoimenovskihtec: too many classes is never a thing ;) The point is everything to be well structured adn to satisfy the S (single responsibility principle) from SOLID

backend/src/main/java/com/htecgroup/skynest/exception/CustomException.java
```diff
@@ -0,0 +1,13 @@
+package com.htecgroup.skynest.exception;
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import org.springframework.http.HttpStatus;
+
+@Getter
+@AllArgsConstructor
+public class CustomException extends RuntimeException {
```
petarstoimenovskihtec: instead of CustomException, you can rename it to SkyNestBaseException or something like that.. it is better than Custom :)

petarstoimenovskihtec (approved):
Just a small suggestion, but anyway I think it looks better right?

backend/src/main/java/com/htecgroup/skynest/exception/CustomException.java
DraganMitrasinovicHtec: SkyNestBaseException has a nice ring to it :D

PR #92, opened 2022-06-23
SJQ2T-135: BE - Tiers, Companies, additional info for Users

ivanfhtec:
* flyway scripts, entities and repositories for companies and tiers

* additional info for users

* added unique index for role names


backend/src/main/java/com/htecgroup/skynest/model/entity/TierEntity.java
```diff
@@ -0,0 +1,28 @@
...
+@Entity(name = "tier")
+@Getter
+@Setter
+@NoArgsConstructor
+@AllArgsConstructor
+public class TierEntity implements Serializable {
+
+  private static final long serialVersionUID = -41839637756579466L;
+
+  public static final String TIER_BASIC = "basic";
```
petarstoimenovskihtec, line 23: this properties should be extracted in some other class, or even an enum

petarstoimenovskihtec: on the other hand, where do you use them?

petarstoimenovskihtec (approved):

backend/src/main/java/com/htecgroup/skynest/model/entity/TierEntity.java
ivanfhtec: nowhere at the moment, we'll need them later

PR #95, opened 2022-06-23
 SJQ2T-80: BE - Create endpoint for password change

MetoMladenovski:
Implemented change password


backend/src/main/java/com/htecgroup/skynest/exception/UserExceptionType.java
```diff
@@ -13,6 +13,8 @@ public enum UserExceptionType {
   INVALID_EMAIL_OR_PASSWORD("The email and password are not valid", HttpStatus.BAD_REQUEST),
   PASSWORDS_DOES_NOT_MATCH(
       "The provided password doesn't match the username", HttpStatus.EXPECTATION_FAILED),
+  OLD_PASSWORD_IS_INCORRECT(
+      "The provided old password is incorrect", HttpStatus.EXPECTATION_FAILED),
```
ivanfhtec, line 17: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/417

we really need to use more meaningful error codes

leave it to @DraganMitrasinovicHtec

ivanfhtec: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/417

we really need to use more meaningful error codes

leave it to @DraganMitrasinovicHtec

ivanfhtec (commented):
looks ok

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -258,4 +259,13 @@ public ResponseEntity<String> deleteUser(@PathVariable UUID uuid) {
     log.info(deleteSuccess);
     return ResponseEntity.ok(deleteSuccess);
   }
+
+  @PostMapping("/password-change/{uuid}")
+  public ResponseEntity<Boolean> changePassword(
+      @Valid @RequestBody UserChangePasswordRequest userChangePasswordRequest,
+      @PathVariable UUID uuid) {
+    userService.authorizeAccessToUserDetailsWith(uuid);
```
DraganMitrasinovicHtec: Should an admin have the power to change other people's passwords? I think this way he can, just wondering if that's correct

ivanfhtec: should be fine :+1: 
I think that's what the client wanted

DarkoVeljanoskiHtec (commented):
looks good to me

backend/src/main/java/com/htecgroup/skynest/exception/UserExceptionType.java
MetoMladenovski: I will change it, I copied it and forgot to change it.

petarstoimenovskihtec: It is better if you don't provide the info that the email is correct but the password is not. If you see all apps respond with **_invalid email or password_** when your credentials are wrong regardless which field is wrong.. But you need to consult with FE and QA team for that change

backend/src/main/java/com/htecgroup/skynest/model/request/UserChangePasswordRequest.java
```diff
@@ -0,0 +1,24 @@
...
+@NoArgsConstructor
+public class UserChangePasswordRequest {
+
+  @NotNull(message = "cannot be null")
+  private String oldPassword;
+
+  @NotNull(message = "cannot be null")
+  @Pattern(
+      regexp =
+          "(?=.*[a-z])(?=.*[A-Z])(?=.*\\d)[a-zA-Z\\d!@#&()–\\[{}\\]:\\-;',?|/*%~$_^+=<>\\s]{8,50}",
```
petarstoimenovskihtec: if same regex is used when we validate the password on register/login, extract it in some constants/pattern class and provide the reference instead

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -121,4 +122,29 @@ public void authorizeAccessToUserDetailsWith(UUID uuid) {
       throw new UserException("Access denied", HttpStatus.FORBIDDEN);
     }
   }
+
+  public void authorizeAccessForChangePassword(UUID uuid) {
+    LoggedUserDto loggedUserDto = currentUserService.getLoggedUser();
```
petarstoimenovskihtec, line 118: There will be a change for this

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -121,4 +122,29 @@ public void authorizeAccessToUserDetailsWith(UUID uuid) {
...
+    }
+  }
+
+  @Override
+  public void changePassword(UserChangePasswordRequest userChangePasswordRequest, UUID uuid) {
+    UserEntity userEntity =
+        userRepository
+            .findById(uuid)
+            .orElseThrow(() -> new UserException(UserExceptionType.USER_NOT_FOUND));
+    if (!bCryptPasswordEncoder.matches(
```
petarstoimenovskihtec: Because this logic exists, try to extract it in one place

MetoMladenovski: The findById?

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
petarstoimenovskihtec: maybe you need to ask again. It is really strange to give somebody power to change your password.
Honestly if that is real app behavior , I would never use that app :)

backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
petarstoimenovskihtec: the findBy -> can go into service method, and the bCryptPasswordEncoder can go in separate service

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
MetoMladenovski: Any user, admin or otherwise can change their own password. We'll leave it like that and when we can ask the client how it should be, we'll change it if we have to.

backend/src/main/java/com/htecgroup/skynest/exception/auth/UnableToChangeOtherUserPasswordException.java
```diff
@@ -0,0 +1,12 @@
+package com.htecgroup.skynest.exception.auth;
+
+import org.springframework.http.HttpStatus;
+
+public class UnableToChangeOtherUserPasswordException extends AuthException {
```
petarstoimenovskihtec: Maybe rename to PasswordChangeForbiddenException.. the message inside tells you the reason

backend/src/main/java/com/htecgroup/skynest/service/UserService.java
```diff
@@ -23,4 +24,10 @@ public interface UserService {
   UserResponse editUser(UserEditRequest userEditRequest, UUID uuid);
 
   void authorizeAccessToUserDetailsWith(UUID uuid);
+
+  void changePassword(UserChangePasswordRequest userChangePasswordRequest, UUID uuid);
```
petarstoimenovskihtec, line 28: Should this methods go in AuthService?

petarstoimenovskihtec (approved):
Few comments, but looks ok

backend/src/main/java/com/htecgroup/skynest/service/UserService.java
MetoMladenovski: I guess editUser is here, it makes sense for change password to be here too..

PR #99, opened 2022-06-24
SJQ2T-199 and SJQ2T-192: BE - Added entities and flyway scripts for files, folders and buckets

ivanfhtec:
diagram

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/model/entity/ObjectEntity.java
+++ b/backend/src/main/java/com/htecgroup/skynest/model/entity/ObjectEntity.java
@@ -1,0 +1,29 @@
+package com.htecgroup.skynest.model.entity;
+
+import javax.persistence.*;
+import java.io.Serializable;
+import java.time.LocalDateTime;
+import java.util.UUID;
+
+@Entity(name = "object")
+@Inheritance(strategy = InheritanceType.JOINED)
+public class ObjectEntity implements Serializable {
+
+  private static final long serialVersionUID = 145843022185009528L;
+
+  @Id @GeneratedValue private UUID id;
+
+  @Column(insertable = false)
+  private LocalDateTime createdOn;
+
+  @Column(insertable = false)
+  private LocalDateTime modifiedOn;
+
+  private LocalDateTime deletedOn;
+
+  @ManyToOne
+  @JoinColumn(name = "created_by", nullable = false)
+  private UserEntity createdBy;
+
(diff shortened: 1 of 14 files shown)


backend/src/main/java/com/htecgroup/skynest/model/entity/ObjectEntity.java
```diff
@@ -0,0 +1,29 @@
...
+@Entity(name = "object")
+@Inheritance(strategy = InheritanceType.JOINED)
+public class ObjectEntity implements Serializable {
+
+  private static final long serialVersionUID = 145843022185009528L;
+
+  @Id @GeneratedValue private UUID id;
+
+  @Column(insertable = false)
+  private LocalDateTime createdOn;
```
petarstoimenovskihtec, line 26: You can use  @CreatedDate and @LastModifiedDate for this audition tables, just you need to allow EntityListeners in your project.

The other option is to use 
    @PrePersist
    @PreUpdate
    @PreRemove

ivanfhtec: I was thinking of doing the same thing we did for companies. We had the database in charge these fields:
<code> `created_on` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP,</code>
<code> `modified_on` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,</code>
But we had to set <code>deleted_on</code> manually.

Would these annotations interfere with that?

ivanfhtec: Or rather, is it better to deal with dates from Java rather than let the database handle it?

ivanfhtec:
I've added the corresponding flyway scripts and made a couple of changes to the entities.

Note that the bucket size field represents the actual size of the bucket on disk. Sending files to trash does not update bucket size, but deleting files does.

backend/src/main/java/com/htecgroup/skynest/model/entity/ObjectEntity.java
petarstoimenovskihtec: well you can leave the database to take care of it, the point is that if you override in Java with lets say some "wrong" value, then database will use those values..  I really leave it up to you

petarstoimenovskihtec (approved):

PR #102, opened 2022-06-27
SJQ2T-118: BE - Endpoint for disable user

MetoMladenovski:
Implemented disable user functionality, added and fixed junit tests.


backend/src/main/java/com/htecgroup/skynest/service/impl/AuthServiceImpl.java
```diff
@@ -74,12 +74,18 @@ public String confirmEmail(String token) {
   @Override
   public boolean isActive(String email) {
     UserDto userDto = userService.findUserByEmail(email);
-    return userDto.getEnabled() && userDto.getVerified() && userDto.getDeletedOn() == null;
```
ivanfhtec, line 77: Splitting this up makes sense.

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
```diff
@@ -53,7 +54,9 @@ public Authentication attemptAuthentication(
       if (loginAttemptService.hasTooManyAttempts(userDetails.getUsername()))
         throw new TooManyAttemptsException();
 
-      if (!authService.isActive(userDetails.getUsername())) throw new EmailNotVerifiedException();
+      if (!authService.isVerified(userDetails.getUsername())) throw new EmailNotVerifiedException();
+
+      if (!authService.isActive(userDetails.getUsername())) throw new UserNotActiveException();
```
ivanfhtec, line 59: is this revealing too much info? should we just return bad credentials or something?

backend/src/main/java/com/htecgroup/skynest/exception/auth/UserNotVerifiedException.java
```diff
@@ -0,0 +1,12 @@
+package com.htecgroup.skynest.exception.auth;
+
+import org.springframework.http.HttpStatus;
+
+public class UserNotVerifiedException extends AuthException {
+
+  public static final String MESSAGE = "Not verified user can't be enabled.";
```
ivanfhtec: unverified

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -352,4 +352,70 @@ public ResponseEntity<Boolean> changePassword(
...
+                    @ExampleObject(
+                        value =
+                            "{\"messages\":[\"User is already disabled.\"],"
+                                + " \"status\": \"409\","
+                                + " \"timestamp\": \"2022-06-07 16:18:12\"}")
+                  })
+            })
+      })
+  @PreAuthorize("hasAuthority(T(com.htecgroup.skynest.model.entity.RoleEntity).ROLE_ADMIN)")
+  @PutMapping("/disable/{uuid}")
```
ivanfhtec: will we always have operations before ids?

in other words, /users/disable/{uuid} instead of /users/{uuid}/disable

not sure which one makes more sense

MetoMladenovski: For me, this way is more natural

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
MetoMladenovski: My logic was, the user should know when he needs to confirm his email to log in, and in the other case when he is banned, he should know there's nothing he can do.

ivanfhtec: makes sense, although most websites would probably send an email notification or something if you try to use a banned account

but this is probably fine for our use case

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
markobarackovhtec: /users/{uuid}/disable you should go with this, this is actually by REST

MetoMladenovski: Okay then, i'll make the change.

backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthenticationFilter.java
markobarackovhtec: This is actually topic for client, you should ask client what she wants here

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
markobarackovhtec: maybe is better to be like this /users/{userId}/disable
this is much more readable

markobarackovhtec: So this was my point, after some resource it goes resource id, so you never gonna do something like this 
/users/function/{userId}/{functionId}, correct way is /users/{userId}/functions/{functionId}

MetoMladenovski: Ahhh, yes, it's way more clear this way.

MetoMladenovski:
The commit name [implemented promote user functionality.] is a mistake!
Should state, changed the url in UserController.

backend/src/main/java/com/htecgroup/skynest/exception/auth/UserNotVerifiedException.java
markobarackovhtec: An unverified user can't be enabled

markobarackovhtec (approved):

PR #107, opened 2022-06-27
SJQ2T-97:BE Endpoint for add company

DraganMitrasinovicHtec:
Admins can now add companies via this endpoint
Using CompanyIO instead of CompanyRequest and CompanyResponse as they would look pretty much the same, can change that if we want to keep consistent tho.


backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
```diff
@@ -42,6 +42,8 @@ protected void configure(HttpSecurity http) throws Exception {
         .authorizeRequests()
         .antMatchers(String.format("%s/**", UrlUtil.PUBLIC_CONTROLLER_URL))
         .permitAll()
+        .antMatchers(String.format("%s/**", UrlUtil.COMPANY_CONTROLLER_URL))
+        .permitAll()
```
ivanfhtec: a leftover from testing?

backend/src/main/java/com/htecgroup/skynest/model/dto/CompanyDto.java
```diff
@@ -12,7 +13,10 @@ public class CompanyDto implements Serializable {
   private LocalDateTime createdOn;
   private LocalDateTime modifiedOn;
   private LocalDateTime deletedOn;
+  private String pib;
```
ivanfhtec, line 15: is CompanyDto used anywhere?

backend/src/main/java/com/htecgroup/skynest/controller/CompanyController.java
```diff
@@ -0,0 +1,168 @@
...
+                            + " \"pib\": \"123456789\","
+                            + "  \"name\": \"Name\","
+                            + "  \"tierName\": \"basic\","
+                            + "  \"phoneNumber\": \"38166575757\","
+                            + "  \"address\": \"Local address\"}")
+              }))
+  @PreAuthorize("hasAuthority(T(com.htecgroup.skynest.model.entity.RoleEntity).ROLE_ADMIN)")
+  @PostMapping
+  public ResponseEntity<CompanyIO> addCompany(@Valid @RequestBody CompanyIO companyIO) {
+
```
petarstoimenovskihtec, line 169: Maybe add a log statement that company is created

backend/src/main/java/com/htecgroup/skynest/controller/CompanyController.java
```diff
@@ -0,0 +1,168 @@
...
+                        "{\"email\": \"username@gmail.com\","
+                            + " \"pib\": \"123456789\","
+                            + "  \"name\": \"Name\","
+                            + "  \"tierName\": \"basic\","
+                            + "  \"phoneNumber\": \"38166575757\","
+                            + "  \"address\": \"Local address\"}")
+              }))
+  @PreAuthorize("hasAuthority(T(com.htecgroup.skynest.model.entity.RoleEntity).ROLE_ADMIN)")
+  @PostMapping
+  public ResponseEntity<CompanyIO> addCompany(@Valid @RequestBody CompanyIO companyIO) {
```
petarstoimenovskihtec: IO suffix usually means that you will do some interaction with file system not that you are referring to a model :D  You can use one model for request, other for response and a parent model that both will extend.. I am sure that there will be other attributes that are only relevant to response object, like id, createdAt, updatedAt etc etc

petarstoimenovskihtec: Does this validation work since you don't have Validated annotation on class level?

backend/src/main/java/com/htecgroup/skynest/service/impl/CompanyServiceImpl.java
```diff
@@ -0,0 +1,49 @@
...
+public class CompanyServiceImpl implements CompanyService {
+
+  private CompanyRepository companyRepository;
+  private TierRepository tierRepository;
+  private ModelMapper modelMapper;
+
+  @Override
+  public CompanyIO addCompany(CompanyIO companyIO) {
+
+    if (companyRepository.existsByPib(companyIO.getPib())) {
```
petarstoimenovskihtec: I would extract this in custom annotation since you already have that

backend/src/main/java/com/htecgroup/skynest/model/dto/CompanyDto.java
```diff
@@ -12,7 +13,10 @@ public class CompanyDto implements Serializable {
...
   private String name;
   private String address;
-  private String domain;
+  private String phoneNumber;
+  private String email;
+  private TierEntity tier;
```
petarstoimenovskihtec: Entity should never be exposed via network.. Maybe create new model such as TierDto or something like that and extract just the attributes that are relevant

backend/src/main/java/com/htecgroup/skynest/security/WebSecurity.java
DraganMitrasinovicHtec: Yep

backend/src/main/java/com/htecgroup/skynest/model/dto/CompanyDto.java
DraganMitrasinovicHtec: It isn't, It was created before so I just left it for possible future use. Should I just delete it for now?

backend/src/main/java/com/htecgroup/skynest/controller/CompanyController.java
DraganMitrasinovicHtec: Of course, forgot about it somehow

backend/src/main/java/com/htecgroup/skynest/controller/CompanyController.java
DraganMitrasinovicHtec: Will do

DraganMitrasinovicHtec: Works, @Valid in the list of arguments is enough.

backend/src/main/java/com/htecgroup/skynest/service/impl/CompanyServiceImpl.java
DraganMitrasinovicHtec: @petarstoimenovskihtec made email and phone number check into an annotation, is there any way to use the same one for user, as Users and Companies use a different repository.

petarstoimenovskihtec: sure, I believe that there is already committed code with custom annotation on service layer. You can take a look how it is made and do similar logic there.
Probably this line `companyRepository.existsByPib(companyIO.getPib()` will go into the validator and here in the service method the code will run only if that annotation satisfy the expression given.

backend/src/main/java/com/htecgroup/skynest/service/impl/CompanyServiceImpl.java
```diff
@@ -0,0 +1,42 @@
...
+
+@Service
+@Validated
+@AllArgsConstructor
+public class CompanyServiceImpl implements CompanyService {
+
+  private CompanyRepository companyRepository;
+  private TierRepository tierRepository;
+  private ModelMapper modelMapper;
+
```
DraganMitrasinovicHtec, line 27: @petarstoimenovskihtec  Something like this?

backend/src/main/java/com/htecgroup/skynest/annotation/validator/UniqueCompanyValidator.java
```diff
@@ -0,0 +1,36 @@
...
+import javax.validation.ConstraintValidator;
+import javax.validation.ConstraintValidatorContext;
+
+@Component
+@AllArgsConstructor
+public class UniqueCompanyValidator
+    implements ConstraintValidator<UniqueCompany, CompanyAddRequest> {
+
+  private final CompanyRepository companyRepository;
+
```
DraganMitrasinovicHtec, line 21: @petarstoimenovskihtec Is throwing exceptions here too bad a practice? I know I should be returning false, but that throws a ConstraintViolationException, which we are already handling as 403-FORBIDDEN.

petarstoimenovskihtec: ConstraintViolationException should not be handled as 403 error, it is always 400 error, because it validates the business logic. If you are handling it as 403, then you have some other problem.
Validator should just return true/false in my opinion

petarstoimenovskihtec: I don't like to block you now, but after merge, consider fixing it

backend/src/main/java/com/htecgroup/skynest/service/impl/CompanyServiceImpl.java
petarstoimenovskihtec: what did we talk here that I need to look ?

petarstoimenovskihtec (approved):

backend/src/main/java/com/htecgroup/skynest/service/impl/CompanyServiceImpl.java
DraganMitrasinovicHtec: The custom annotation

backend/src/main/java/com/htecgroup/skynest/annotation/validator/UniqueCompanyValidator.java
DraganMitrasinovicHtec: Makes sense, will change it when we fix ConstraintViolationException handling.

PR #108, opened 2022-06-27
SJQ2T-197:  BE - Endpoint for view bucket details

DarkoVeljanoskiHtec:
added endpoint in the controller, service and implemented the functionality for view bucket details, also added test and necessary utils for the test classes


backend/src/main/java/com/htecgroup/skynest/model/response/BucketResponse.java
```diff
@@ -0,0 +1,18 @@
+package com.htecgroup.skynest.model.response;
+
+import com.htecgroup.skynest.model.entity.CompanyEntity;
+import com.htecgroup.skynest.model.entity.UserEntity;
+import lombok.Data;
+
+@Data
+public class BucketResponse {
+  private UserEntity createdBy;
```
ivanfhtec: should we send entire entities in responses?

or just emails/ids

backend/src/main/java/com/htecgroup/skynest/model/response/BucketResponse.java
```diff
@@ -0,0 +1,18 @@
...
+
+  private String name;
+
+  private CompanyEntity company;
```
ivanfhtec: same as line 9

backend/src/main/java/com/htecgroup/skynest/model/response/BucketResponse.java
petarstoimenovskihtec: As I wrote in the other PR, entities should never be exposed over the network.. create separate model and add just the necessary attributes

backend/src/test/java/com/htecgroup/skynest/service/impl/BucketServiceImplTest.java
```diff
@@ -0,0 +1,41 @@
...
+import java.util.Optional;
+
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class BucketServiceImplTest {
+
+  @Mock private BucketRepository bucketRepository;
+  @Spy private ModelMapper modelMapper;
```
petarstoimenovskihtec, line 25: why Spy ? didn't we already have how to mock this modelMapper in other tests?

backend/src/test/java/com/htecgroup/skynest/utils/BucketEntityUtil.java
```diff
@@ -0,0 +1,10 @@
+package com.htecgroup.skynest.utils;
+
+import com.htecgroup.skynest.model.entity.BucketEntity;
+
+public class BucketEntityUtil extends BucketBasicUtil {
+
+  public static BucketEntity getNotPublic() {
```
petarstoimenovskihtec: why not rename to getPrivateBucket?

backend/src/test/java/com/htecgroup/skynest/utils/BucketBasicUtil.java
```diff
@@ -0,0 +1,15 @@
+package com.htecgroup.skynest.utils;
+
+import com.htecgroup.skynest.model.entity.CompanyEntity;
+
+import java.time.LocalDateTime;
+
+public class BucketBasicUtil {
+  protected static final LocalDateTime currentDateTime = LocalDateTime.now();
```
petarstoimenovskihtec: I would not extract current date in a util class since that is created once at the moment of creation and not when you actually call it..

On the other hand, util classes can be final and with private constructor (I also believe that we have that already in the code)

backend/src/test/java/com/htecgroup/skynest/utils/BucketBasicUtil.java
```diff
@@ -0,0 +1,15 @@
...
+  protected static final UserEntityUtil userEntity = new UserEntityUtil();
+  protected static final CompanyEntity companyEntity = new CompanyEntity();
+  protected static String name = "Name";
+  protected static String description = "Description";
+  protected static long size = 100;
```
petarstoimenovskihtec: you can use wrapper instead.. Long

backend/src/test/java/com/htecgroup/skynest/utils/BucketBasicUtil.java
```diff
@@ -0,0 +1,15 @@
...
+  protected static boolean isPublic = false;
```
petarstoimenovskihtec: hmm. when we call this variable we do not know that the value is false.. maybe rename it to something that will tell us that the bucket is not public :) like privateBucket

backend/src/test/java/com/htecgroup/skynest/utils/BucketBasicUtil.java
```diff
@@ -0,0 +1,15 @@
+package com.htecgroup.skynest.utils;
+
+import com.htecgroup.skynest.model.entity.CompanyEntity;
+
+import java.time.LocalDateTime;
+
+public class BucketBasicUtil {
+  protected static final LocalDateTime currentDateTime = LocalDateTime.now();
+  protected static final UserEntityUtil userEntity = new UserEntityUtil();
```
petarstoimenovskihtec: Why do you create object from util class in util class?? Why not using it directly where you needed it? Anyway you should not be able to create util object since all the methods all static there and constructor is hidden aka private

backend/src/main/java/com/htecgroup/skynest/controller/BucketsController.java
```diff
@@ -0,0 +1,105 @@
...
+      value = {
+        @ApiResponse(
+            responseCode = "200",
+            description = "Bucket returned",
+            content = {
+              @Content(
+                  mediaType = "application/json",
+                  schema = @Schema(implementation = UserResponse.class),
+                  examples = {
+                    @ExampleObject(
```
MarkoRadunovicHtec: Is this wrong example object for status code 200? Should be BucketResponse.class and corresponding example. Not UserResponse example.

backend/src/main/java/com/htecgroup/skynest/model/response/BucketResponse.java
DarkoVeljanoskiHtec: You're right, I'll make the change

backend/src/test/java/com/htecgroup/skynest/service/impl/BucketServiceImplTest.java
DarkoVeljanoskiHtec: In every test class we have it's annotated with Spy, should we change it ?

petarstoimenovskihtec: the ModelMapper ??

DarkoVeljanoskiHtec: yes the ModelMapper in the tests classes.

backend/src/main/java/com/htecgroup/skynest/controller/BucketsController.java
DarkoVeljanoskiHtec: Yes, you're right I forgot to change the example after copying it from the other class.

backend/src/test/java/com/htecgroup/skynest/utils/BucketBasicUtil.java
DarkoVeljanoskiHtec: I was just following the foundation set by userEntityUtil, I'll try to change it

backend/src/test/java/com/htecgroup/skynest/service/impl/BucketServiceImplTest.java
petarstoimenovskihtec: well yes

backend/src/test/java/com/htecgroup/skynest/utils/BucketBasicUtil.java
```diff
@@ -0,0 +1,11 @@
+package com.htecgroup.skynest.utils;
+
+import com.htecgroup.skynest.model.entity.CompanyEntity;
+
+public class BucketBasicUtil {
+  protected static final CompanyEntity companyEntity = new CompanyEntity();
+  protected static String name = "Name";
+  protected static String description = "Description";
+  protected static Long size = Long.valueOf(1000);
```
petarstoimenovskihtec: 1000L

backend/src/test/java/com/htecgroup/skynest/utils/BucketEntityUtil.java
```diff
@@ -0,0 +1,10 @@
+package com.htecgroup.skynest.utils;
+
+import com.htecgroup.skynest.model.entity.BucketEntity;
+
+public class BucketEntityUtil extends BucketBasicUtil {
+
```
petarstoimenovskihtec, line 15: I am wondering, if you just extend this BucketBasicUtil for property access, why don't you move them inside this class, and remove the BucketBasicUtil.. 

Also make this class final and with private constructor, because that is the usage of Util class, not extendable and closed for creating objects

backend/src/test/java/com/htecgroup/skynest/service/impl/BucketServiceImplTest.java
DarkoVeljanoskiHtec: If it's not Spy, then it throws null pointer exception, if we mock it with Mock then we need to use Model mapper in almost every test, so this was the idea by @MetoMladenovski for excluding Model mapper from every test refactor.

petarstoimenovskihtec: Didn't Meto already exclude that for the tests, and use the "mapped" object instead?

MetoMladenovski: I am not sure I understand, but I did leave it as a Spy ModelMapper, because we use it too much in our service and all it does is map the Entity to Response,Dto etc. 
If I mock the ModelMapper we would have to mock all those .map() methods in the services. My logic was to use Spy to not get the NullPointerException, and have modelMapper map those objects.

backend/src/test/java/com/htecgroup/skynest/utils/BucketEntityUtil.java
DarkoVeljanoskiHtec: make sense, I've changed it like that

backend/src/test/java/com/htecgroup/skynest/service/impl/BucketServiceImplTest.java
petarstoimenovskihtec: ok lets leave it like that for the time being, but Spy is not the way to do it :)

petarstoimenovskihtec (approved):

PR #112, opened 2022-06-28
SJQ2T-117: BE - Endpoint for enable user

MetoMladenovski:
Implemented enable user functionality.


backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -158,4 +159,17 @@ public void disableUser(UUID userId) {
...
+  @Override
+  public void enableUser(UUID uuid) {
+    UserDto userDto = findUserById(uuid);
+    if (!userDto.getVerified()) {
+      throw new UserNotVerifiedException();
+    }
+    if (userDto.getEnabled() && userDto.getDeletedOn() == null) {
+      throw new UserAlreadyEnabledException();
+    }
+    UserDto enabledUserDto = userDto.withEnabled(true).withDeletedOn(null);
```
petarstoimenovskihtec: you can create a method in userDto that will contain this value setting. It will be more readable and will give sense what is done, maybe enableUser() is pretty descriptive or use your imagination

petarstoimenovskihtec (approved):
When you apply the changes, merge the code

PR #114, opened 2022-06-28
SQJ2T-223: BE - Action service

ivanfhtec:
* added action service

Should I add some filtering by date (performedOn) as well? What kind? Or should we do that when necessary?


DarkoVeljanoskiHtec (approved):
its good for now

PR #117, opened 2022-06-29
SJQ2T-196: BE Endpoint for View Buckets

DarkoVeljanoskiHtec:
endpoint, service and test for view bucket functionality.


backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
```diff
@@ -97,4 +98,60 @@ public ResponseEntity<BucketResponse> getBucket(@PathVariable UUID uuid) {
...
+                            "[{\"id\": \"a6fd6d95-0a60-43ff-961f-2b9b2ff72f95\","
+                                + "\"name\": \"Name\","
+                                + "\"companyId\": \"a6fd6d95-0a60-43ff-961f-2b9b2ff72f95\","
+                                + "\"description\": \"Description\","
+                                + "\"size\": \"1000\"}},"
+                                + "{\"id\": \"u7yd987h-0a79-42dd-961s-7sfh564kdv2s\","
+                                + "\"name\": \"Name\","
+                                + "\"companyId\": \"a6fd6d95-0a60-43ff-961f-2b9b2ff72f95\","
+                                + "\"description\": \"Description\","
+                                + "\"size\": \"1000\"}]")
```
ivanfhtec: two examples objects that are exactly the same?

backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
```diff
@@ -97,4 +98,60 @@ public ResponseEntity<BucketResponse> getBucket(@PathVariable UUID uuid) {
         new ResponseEntity<>(bucketResponse, HttpStatus.OK);
     return bucketResponseEntity;
   }
+
+  @Operation(summary = "Get all buckets")
```
ivanfhtec, line 173: should this get all buckets just for the time being, or should we only get the user's private buckets as well as all public buckets from the user's company?

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
```diff
@@ -25,4 +27,12 @@ public BucketResponse getBucket(UUID uuid) {
     BucketResponse bucketResponse = modelMapper.map(bucketEntity, BucketResponse.class);
     return bucketResponse;
   }
+
+  @Override
+  public List<BucketResponse> listAllBuckets() {
+    List<BucketEntity> entityList = (List<BucketEntity>) bucketsRepository.findAll();
```
ivanfhtec: findAll() returns an Iterable? I can't believe this works

nice

backend/src/test/java/com/htecgroup/skynest/service/impl/BucketServiceImplTest.java
```diff
@@ -25,6 +27,21 @@ class BucketServiceImplTest {
   @Spy private ModelMapper modelMapper;
   @Spy @InjectMocks private BucketServiceImpl bucketService;
 
+  @Test
+  void listAllBuckets() {
+    List<BucketEntity> bucketEntityList = new ArrayList<>();
+    bucketEntityList.add(BucketEntityUtil.getPrivateBucket());
```
ivanfhtec: Collections.singletonList()?

backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
DarkoVeljanoskiHtec: I forgot to change it

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
DarkoVeljanoskiHtec: Yes, it works like this :)

backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
DarkoVeljanoskiHtec: I don't know, I thought we're going to do that later with the permissions, but we could talk it out.

ivanfhtec (approved):

PR #118, opened 2022-06-29
SJQ2T-112: BE - Endpoint for demoting user

MetoMladenovski:
Implemented demote user functionality


backend/src/main/java/com/htecgroup/skynest/annotation/OnlyManagerCanBeDemoted.java
```diff
@@ -0,0 +1,21 @@
...
+import javax.validation.Payload;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ElementType.PARAMETER})
+@Constraint(validatedBy = DemoteValidator.class)
+public @interface OnlyManagerCanBeDemoted {
```
petarstoimenovskihtec: If you have a annotation that is checking if user is manager, use that.
If not since I am not sure if you have different annotations for checking if the user is admin, worker or manager.. you can combine that in one annotation and use it with different input params.
Try to make the annotation reusable.. check with @MarkoRadunovicHtec , I already showed him how to do it

MetoMladenovski: Marko helped me, and I think we did it.

DarkoVeljanoskiHtec (approved):
Looks good

PR #120, opened 2022-06-29
SJQ2T-224: BE - Admin should not be able to disable another admin 

MetoMladenovski:
We had a bug where admin could disable and enable another admin, so this is the fix.

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
+++ b/backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
@@ -145,7 +145,7 @@ public class UserServiceImpl implements UserService {
   }
 
   @Override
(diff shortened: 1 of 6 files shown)


backend/src/main/java/com/htecgroup/skynest/service/impl/UserServiceImpl.java
```diff
@@ -148,7 +145,7 @@ public void changePassword(UserChangePasswordRequest userChangePasswordRequest,
   }
```
DraganMitrasinovicHtec, line 146: Since this basically checks if the User with the given uuid is an admin, maybe its better if it's called @NotAdmin or something similar?

MetoMladenovski: You are right, yes.

petarstoimenovskihtec (approved):

PR #123, opened 2022-06-29
SJQ2T-195: BE Endpoint For Delete Bucket

DarkoVeljanoskiHtec:
added endpoint in the controller, implemented the service and added two tests as well.

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
+++ b/backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
@@ -11,8 +11,8 @@ import io.swagger.v3.oas.annotations.responses.ApiResponses;
 import lombok.AllArgsConstructor;
 import lombok.extern.log4j.Log4j2;
 import org.springframework.http.ResponseEntity;
-import org.springframework.web.bind.annotation.DeleteMapping;
 import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.PutMapping;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RestController;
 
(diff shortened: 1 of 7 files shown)


backend/src/main/java/com/htecgroup/skynest/service/BucketService.java
```diff
@@ -1,10 +1,15 @@
...
+import com.htecgroup.skynest.model.dto.BucketDto;
 import com.htecgroup.skynest.model.response.BucketResponse;
 
 import java.util.UUID;
 
 public interface BucketService {
 
   BucketResponse getBucket(UUID uuid);
+
+  BucketDto findBucketById(UUID uuid);
```
petarstoimenovskihtec, line 15: What is the correct returned mode? you have BucketResponse and BucketDto ? try to be consistent as in the other services

DarkoVeljanoskiHtec: BucketResponse is the correct one who is connected to the controller, but I use BucketDto for the delete implementation and for easier tests, the same thing is done in the user service as well.

backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
```diff
@@ -97,4 +94,54 @@ public ResponseEntity<BucketResponse> getBucket(@PathVariable UUID uuid) {
...
+                    @ExampleObject(
+                        value =
+                            "{\"messages\":[\"Bucket is already disabled.\"],"
+                                + " \"status\": \"409\","
+                                + " \"timestamp\": \"2022-06-07 16:18:12\"}")
+                  })
+            })
+      })
+  @PutMapping("delete/{uuid}")
+  public ResponseEntity<Boolean> deleteBucket(@PathVariable UUID uuid) {
```
ivanfhtec, line 231: can this return 401?

backend/src/main/java/com/htecgroup/skynest/service/BucketService.java
ivanfhtec: Could we use BucketEntity here instead then?

Other services wouldn't have to import bucket repo

More refactoring

backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
DarkoVeljanoskiHtec: We don't have preAuthorize yet so i guess not.

ivanfhtec: invalid jwt? this endpoint requires a token, I think

ivanfhtec (approved):

PR #124, opened 2022-06-29
SJQ2T-98: BE - Endpoint for Delete company

DraganMitrasinovicHtec:
Added an endpoint for deleting companies. It's a soft delete, just updating the deletedOn field to the current date.

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/controller/CompanyController.java
+++ b/backend/src/main/java/com/htecgroup/skynest/controller/CompanyController.java
@@ -16,12 +16,10 @@ import lombok.extern.log4j.Log4j2;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
 import org.springframework.security.access.prepost.PreAuthorize;
-import org.springframework.web.bind.annotation.PostMapping;
-import org.springframework.web.bind.annotation.RequestBody;
-import org.springframework.web.bind.annotation.RequestMapping;
-import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.bind.annotation.*;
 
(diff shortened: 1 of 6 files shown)


backend/src/main/java/com/htecgroup/skynest/controller/CompanyController.java
```diff
@@ -170,4 +168,71 @@ public ResponseEntity<CompanyResponse> addCompany(
...
+              @Content(
+                  mediaType = "application/json",
+                  schema = @Schema(implementation = String.class),
+                  examples = {@ExampleObject(value = "Internal Server Error")})
+            })
+      })
+  @PreAuthorize("hasAuthority(T(com.htecgroup.skynest.model.entity.RoleEntity).ROLE_ADMIN)")
+  @DeleteMapping("/{uuid}")
+  @ResponseStatus(value = HttpStatus.OK)
+  public void deleteCompany(@PathVariable UUID uuid) {
```
ivanfhtec, line 234: void?

DraganMitrasinovicHtec: Do we really need to return string for delete operations? You can tell if it's successful by the status code?

MetoMladenovski: ResponseEntity(Boolean) i think is the best

ivanfhtec: either one is fine, as long as we're consistent

DraganMitrasinovicHtec: No prob, will change

PR #125, opened 2022-06-29
SJQ2T-228: BE - Files and Folders must have unique names

ivanfhtec:
* added triggers to enforce unique names for files and folders (inside of a folder)

* generic errors now have all causes displayed

This doesn't cover bucket names. Should they be unique? Since they are shared between a bunch of people, how would that work?

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthorizationFilter.java
+++ b/backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthorizationFilter.java
@@ -22,6 +22,8 @@ import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 
 @Log4j2
 @AllArgsConstructor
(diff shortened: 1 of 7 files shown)


backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthorizationFilter.java
```diff
@@ -80,11 +82,20 @@ protected void doFilterInternal(
       ExceptionUtil.writeToResponse(errorMessage, response);
     } catch (Exception ex) {
       log.error(ex);
+
+      List<String> messages = new ArrayList<>();
+      Throwable currentException = ex;
+      while (currentException != null && currentException != currentException.getCause()) {
```
petarstoimenovskihtec, line 88: Can you tell me what is the idea here?

petarstoimenovskihtec: you are trying to get the first error in this nested loop ?

ivanfhtec: The earlier version of this wouldn't show errors coming from any of the triggers, since those errors are a cause of a cause of a cause...

Now, users get the whole chain of causes.

ivanfhtec:
As it stands, these triggers enforce unique names even for deleted objects.

PR #128, opened 2022-06-29
SJQ2T-229: BE - Basic file upload/download

ivanfhtec:
* no tests

* no documentation

* no request validation

* no logs

seems legit


backend/src/main/java/com/htecgroup/skynest/repository/FileMetadataRepository.java
```diff
@@ -0,0 +1,13 @@
...
+
+import com.htecgroup.skynest.model.entity.BucketEntity;
+import com.htecgroup.skynest.model.entity.FileMetadataEntity;
+import com.htecgroup.skynest.model.entity.FolderEntity;
+import org.springframework.data.repository.CrudRepository;
+
+import java.util.UUID;
+
+public interface FileMetadataRepository extends CrudRepository<FileMetadataEntity, UUID> {
+  boolean existsByNameAndBucketAndParentFolder(
```
DraganMitrasinovicHtec, line 11: This checks for a unique combination of these 3 attributes, like a composite key?

ivanfhtec: and it works somehow

[the db will also have triggers for the same thing](https://github.com/JDP2022SkyNest/SkyNest/pull/125)

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -0,0 +1,169 @@
...
+
+  private String storeFileContentsToDatabase(MultipartFile multipartFile) throws IOException {
+    Object objectId =
+        operations.store(
+            multipartFile.getInputStream(),
+            multipartFile.getOriginalFilename(),
+            multipartFile.getContentType());
+    return objectId.toString();
+  }
+
```
DraganMitrasinovicHtec, line 94: The logic here seems a little convoluted, lots of helper functions. Mentors would probably ask for the checks to be removed to a separate class (or a custom annotation :)), but I can live with this.

DraganMitrasinovicHtec (dismissed):
Rough around the edges, but good for starters

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
ivanfhtec: those helper functions are supposed to make this easier to read

not sure if it worked though

DarkoVeljanoskiHtec: I agree with this, it's kinda hard to read it like this, but maybe it's good practice like this I don't know.

DraganMitrasinovicHtec: Maybe it would be okay if the checks are somewhere else.

ivanfhtec:
now has docs and logs
also checks available space in buckets and doesn't allow uploads to deleted buckets

DraganMitrasinovicHtec (approved):
Looks great now

PR #131, opened 2022-06-30
SJQ2T-105: BE - Endpoint for add company for user

MetoMladenovski:
implemented add company for user functionality.


backend/src/main/java/com/htecgroup/skynest/service/UserService.java
```diff
@@ -49,4 +49,7 @@ void demoteUser(
           UUID uuid);
 
   void disableUser(@Valid @UserNotAdmin UUID userId);
+
+  void addCompanyForUser(
+      @Valid @UserHasCompany UUID uuid, @Valid @AddUserInAdminCompany UUID companyId);
```
ivanfhtec: so the user must not already be in a company, and admins can only add users to their own company

shouldn't it be something like UserNotInACompany instead of UserHasCompany?

and AdminIsInCompany or CurrentUserCompany instead of AddUserInAdminCompany?

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
```diff
@@ -735,4 +735,13 @@ public ResponseEntity<LoggedUserResponse> getLoggedUser() {
     LoggedUserResponse loggedUser = currentUserService.getUserResponseForLoggedUser();
     return ResponseEntity.ok(loggedUser);
   }
+
+  @PreAuthorize("hasAuthority(T(com.htecgroup.skynest.model.entity.RoleEntity).ROLE_ADMIN)")
+  @PutMapping("/{uuid}/add-company/{companyId}")
+  public ResponseEntity<Boolean> addCompanyForUser(
```
ivanfhtec, line 801: needs more swagger

backend/src/main/java/com/htecgroup/skynest/exception/company/CompanyNotFoundException.java
```diff
@@ -0,0 +1,12 @@
+package com.htecgroup.skynest.exception.company;
+
+import org.springframework.http.HttpStatus;
+
+public class CompanyNotFoundException extends CompanyException {
+
+  public static final String MESSAGE = "Company doesn't exist";
```
ivanfhtec: Company not found

so that all 404s look the same

backend/src/main/java/com/htecgroup/skynest/service/CompanyService.java
```diff
@@ -1,12 +1,16 @@
...
 import com.htecgroup.skynest.model.response.CompanyResponse;
 
 import javax.validation.Valid;
+import java.util.UUID;
 
 public interface CompanyService {
 
   CompanyResponse addCompany(@Valid @UniqueCompany CompanyAddRequest companyAddRequest);
+
+  CompanyDto findById(UUID companyId);
```
ivanfhtec, line 15: we're sticking with dtos?

MetoMladenovski: Unfortunately yes. It's used in too many places

backend/src/main/java/com/htecgroup/skynest/controller/UserController.java
MetoMladenovski: Yeap

backend/src/main/java/com/htecgroup/skynest/service/CompanyService.java
DarkoVeljanoskiHtec: Its too late now...

backend/src/main/java/com/htecgroup/skynest/service/UserService.java
MetoMladenovski: The first one I agree with, the second one you I think it's still missing something, maybe AdminAndUserInSameCompany

DraganMitrasinovicHtec (approved):
Good job, you even wrote tests for everything :D

backend/src/main/java/com/htecgroup/skynest/service/UserService.java
ivanfhtec: the validator checks whether the current user, an admin, is in the company that's in the request

calling it AdminAndUserInSameCompany makes it seem like the user is already in that company, or at least that the annotation checks that

ivanfhtec: based on what the validator checks, CurrentUserIsInCompany probably makes sense

PR #136, opened 2022-07-01
SJQ2T-231: BE - Endpoint For Add Folder

DarkoVeljanoskiHtec:
Added new controller and endpoint, new service and functionality, new response and request classes, as well as few tests and the necessary util classes for them.


backend/src/main/java/com/htecgroup/skynest/model/response/FolderResponse.java
```diff
@@ -0,0 +1,17 @@
...
+import java.util.UUID;
+
+@Data
+public class FolderResponse {
+
+  private UUID createdById;
+
+  private String name;
+
+  private String bucketName;
```
ivanfhtec: bucket names aren't unique, so this has to be bucket id

backend/src/main/java/com/htecgroup/skynest/model/response/FolderResponse.java
```diff
@@ -0,0 +1,17 @@
...
+
+  private String parentFolderName;
```
ivanfhtec: same problem here, why not id?

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
```diff
@@ -0,0 +1,105 @@
...
+            description = "Internal Server Error",
+            content = {
+              @Content(
+                  mediaType = "application/json",
+                  schema = @Schema(implementation = String.class),
+                  examples = {@ExampleObject(value = "Internal Server Error")})
+            })
+      })
+  @PostMapping
+  public ResponseEntity<FolderResponse> createFolder(
```
ivanfhtec, line 94: no 403s? for private buckets at least

backend/src/main/java/com/htecgroup/skynest/repository/BucketRepository.java
```diff
@@ -3,6 +3,9 @@
 import com.htecgroup.skynest.model.entity.BucketEntity;
 import org.springframework.data.repository.CrudRepository;
 
+import java.util.Optional;
 import java.util.UUID;
 
-public interface BucketRepository extends CrudRepository<BucketEntity, UUID> {}
+public interface BucketRepository extends CrudRepository<BucketEntity, UUID> {
+  Optional<BucketEntity> findBucketByName(String name);
```
ivanfhtec, line 11: since name isn't unique. this just finds some bucket with that name

backend/src/main/java/com/htecgroup/skynest/repository/FolderRepository.java
```diff
@@ -0,0 +1,13 @@
...
+import com.htecgroup.skynest.model.entity.FolderEntity;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+
+import java.util.Optional;
+import java.util.UUID;
+
+@Repository
+public interface FolderRepository extends JpaRepository<FolderEntity, UUID> {
+  Optional<FolderEntity> findFolderByName(String name);
```
ivanfhtec: same thing, folder names are not unique

backend/src/main/java/com/htecgroup/skynest/model/dto/FolderDto.java
```diff
@@ -0,0 +1,32 @@
...
+import lombok.NoArgsConstructor;
+import lombok.With;
+
+import java.time.LocalDateTime;
+import java.util.UUID;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class FolderDto {
```
ivanfhtec, line 14: must we have more dtos

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -0,0 +1,66 @@
...
+        || folderCreateRequest.getParentFolderName().isEmpty()) {
+      folderEntity.setParentFolder(null);
+    } else {
+      FolderEntity parentFolderEntity =
+          modelMapper.map(
+              findFolderByName(folderCreateRequest.getParentFolderName()), FolderEntity.class);
+      folderEntity.setParentFolder(parentFolderEntity);
+    }
+
+    folderEntity.setBucket(bucketEntity);
```
ivanfhtec, line 64: could users specify a parent folder that's not actually in this bucket?

assuming you use ids instead of names in the request model

backend/src/main/java/com/htecgroup/skynest/model/dto/FolderDto.java
DarkoVeljanoskiHtec: Well that's how we worked so far, I'm just keeping up consistency.

backend/src/main/java/com/htecgroup/skynest/model/response/FolderResponse.java
DarkoVeljanoskiHtec: I forgot about that, I guess I'll change it.

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
DarkoVeljanoskiHtec: Well considering I take it from the request, it only depends what get sent as parent folder in the request, so yes.

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
DarkoVeljanoskiHtec: sure, but we still don't have the permissions so it's pointless as of now.

backend/src/main/java/com/htecgroup/skynest/repository/BucketRepository.java
ivanfhtec: do you use this method anywhere?

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -0,0 +1,73 @@
...
+    folderEntity.setCreatedBy(currentUser);
+
+    folderEntity = folderRepository.save(folderEntity);
+    return modelMapper.map(folderEntity, FolderResponse.class);
+  }
+
+  @Override
+  public FolderDto findFolderByName(String name) {
```
ivanfhtec: where do you use this method?

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
ivanfhtec: don't forget to check whether the parent folder is in that bucket

or we could just ignore the bucket sent in the request, and just use the parent folder's bucket
if parent folder isn't null

backend/src/main/java/com/htecgroup/skynest/repository/BucketRepository.java
DarkoVeljanoskiHtec: I don't but I left it anyway for future uses, you think I should remove it ?

ivanfhtec: you'd ask for a bucket with a certain name, and it returns pretty much any bucket with that name

I don't see where that would be useful

DarkoVeljanoskiHtec: I keep forgetting that lol, I'll remove it.

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
DarkoVeljanoskiHtec: I did it with custom annotation

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+      parentFolderEntity = folderRepository.findFolderById(folderCreateRequest.getParentFolderId());
+    }
+
+    folderEntity = setNewFolder(folderEntity, currentUser, bucketEntity, parentFolderEntity);
+    folderEntity = folderRepository.save(folderEntity);
+    return modelMapper.map(folderEntity, FolderResponse.class);
+  }
+
+  @Override
+  public FolderEntity setNewFolder(
```
ivanfhtec: is this method public so that we could reuse this code to move folders into other folders?

backend/src/main/java/com/htecgroup/skynest/annotation/validator/ParentFolderIsInTheSameBucketValidator.java
```diff
@@ -0,0 +1,44 @@
...
+      ConstraintValidatorContext constraintValidatorContext) {
+    BucketEntity bucketEntity =
+        modelMapper.map(
+            bucketService.findBucketById(folderCreateRequest.getBucketId()), BucketEntity.class);
+    FolderEntity parentFolderEntity;
+    if (folderCreateRequest.getParentFolderId() != null) {
+      parentFolderEntity = folderRepository.findFolderById(folderCreateRequest.getParentFolderId());
+    } else {
+      return true;
+    }
```
ivanfhtec: rewriting lines 32-37 to this might be a bit easier to read:
>
    
    if (folderCreateRequest.getParentFolderId() == null) {
      return true;
    }
    FolderEntity parentFolderEntity = folderRepository.findFolderById(folderCreateRequest.getParentFolderId());

backend/src/main/java/com/htecgroup/skynest/annotation/validator/ParentFolderIsInTheSameBucketValidator.java
```diff
@@ -0,0 +1,44 @@
...
+
+    if (bucketEntity.getId() != parentFolderEntity.getBucket().getId()) {
```
ivanfhtec, line 42: shouldn't this be .equals()?

backend/src/main/java/com/htecgroup/skynest/annotation/validator/ParentFolderIsInTheSameBucketValidator.java
DarkoVeljanoskiHtec: I like that

backend/src/main/java/com/htecgroup/skynest/annotation/validator/ParentFolderIsInTheSameBucketValidator.java
DarkoVeljanoskiHtec: I don't know, but it works like this :)

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
DarkoVeljanoskiHtec: Well tbh I didn't think about that but why not.

backend/src/main/java/com/htecgroup/skynest/annotation/validator/ParentFolderIsInTheSameBucketValidator.java
petarstoimenovskihtec: if it is Wrapper object then it needs to be equals, if it is primitive simple != is enough

backend/src/main/java/com/htecgroup/skynest/annotation/validator/ParentFolderIsInTheSameBucketValidator.java
petarstoimenovskihtec: basically  
`if (folderCreateRequest.getParentFolderId() == null) { 
      return true; 
 }` 
should be the first line in this method. Because if that is true, then findBucketById is not needed, so you have obsolete call to DB

backend/src/main/java/com/htecgroup/skynest/model/dto/FolderDto.java
```diff
@@ -0,0 +1,32 @@
...
+  private UUID id;
+  private LocalDateTime createdOn;
+
+  private LocalDateTime modifiedOn;
+
+  @With private LocalDateTime deletedOn;
+
+  private UserEntity createdBy;
```
petarstoimenovskihtec: Since you have a DTO, all attributes inside needs to be wrappers, primitives or other DTOs, you have Entity object here so it kinda messed the whole concept

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+      FolderEntity folderEntity,
+      UserEntity currentUser,
+      BucketEntity bucketEntity,
+      FolderEntity parentFolderEntity) {
+    folderEntity.setParentFolder(parentFolderEntity);
```
petarstoimenovskihtec, line 63: I believe that this setter thing we discussed multiple times :)

DarkoVeljanoskiHtec: Should I do this with builder too ?

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -0,0 +1,67 @@
...
+  private ModelMapper modelMapper;
+  private FolderRepository folderRepository;
+  private CurrentUserService currentUserService;
+
+  private UserRepository userRepository;
+
+  private BucketService bucketService;
+
+  @Override
+  public FolderResponse createFolder(
```
petarstoimenovskihtec, line 38: According to me this method needs complete rewriting. It looks too complicated with a lot of things tightly coupled. 
If I understand correctly FolderEntity has a FK to Bucket and User, that means that you need only the ids not the whole objects, since you are not creating neither the buckets nor the user, you just need the reference.

Having that said, 
```
BucketEntity bucketEntity =
        modelMapper.map(
            bucketService.findBucketById(folderCreateRequest.getBucketId()), BucketEntity.class);
```
and 
`UserEntity currentUser = userRepository.getById(loggedUserDto.getUuid());`
are totally unnecessary.

Probably same can be done with `parentFolder` and use just the id, if you already check it in the annotation it self

petarstoimenovskihtec (changes requested):
Some changes

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
petarstoimenovskihtec: yes

DarkoVeljanoskiHtec: This won't work with builder or with annotation, only way this works is with the setters.

petarstoimenovskihtec (approved):

PR #138, opened 2022-07-01
SJQ2T-254: BE - Add position in company to user responses

DraganMitrasinovicHtec:
Everywhere we return a UserResponse I added positionInCompany, changed documentation too.


backend/src/main/java/com/htecgroup/skynest/model/request/UserEditRequest.java
```diff
@@ -30,4 +30,8 @@ public class UserEditRequest {
   @NotBlank(message = "cannot be null or empty")
   @Size(max = 254, message = "length cannot be over 254")
   private String address;
+
+  @NotBlank(message = "cannot be null or empty")
+  @Size(max = 100, message = "length cannot be over 100")
+  private String positionInCompany;
```
ivanfhtec, line 36: I think it's 255 in the db

though it doesn't really matter

DraganMitrasinovicHtec: Let's keep it consistent then

ivanfhtec (approved):

PR #139, opened 2022-07-01
SJQ2T-257: BE - Add Endpoint For Restore Bucket

DarkoVeljanoskiHtec:
Added endpoint for restoring buckets, implemented functionality in the service and added few tests.


backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
```diff
@@ -294,6 +294,56 @@ public ResponseEntity<Boolean> deleteBucket(@PathVariable UUID uuid) {
...
+                            "{\"messages\":[\"Restored is already enabled.\"],"
+                                + " \"status\": \"409\","
+                                + " \"timestamp\": \"2022-06-07 16:18:12\"}")
+                  })
+            })
+      })
+  @PutMapping("/{uuid}/restore")
+  public ResponseEntity<Boolean> restoreBucket(@PathVariable UUID uuid) {
+    bucketService.restoreBucket(uuid);
+    log.info("Bucket with id {} was successfully restored", uuid);
```
ivanfhtec: use the action service instead

so far, we've been recording actions in services rather than controllers

though it's really the same either way

backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
```diff
@@ -294,6 +294,56 @@ public ResponseEntity<Boolean> deleteBucket(@PathVariable UUID uuid) {
...
+                    @ExampleObject(
+                        value =
+                            "{\"messages\":[\"Restored is already enabled.\"],"
+                                + " \"status\": \"409\","
+                                + " \"timestamp\": \"2022-06-07 16:18:12\"}")
+                  })
+            })
+      })
+  @PutMapping("/{uuid}/restore")
+  public ResponseEntity<Boolean> restoreBucket(@PathVariable UUID uuid) {
```
ivanfhtec, line 356: this could also return 401 (invalid token)?

ivanfhtec (approved):
seems good

ivanfhtec:
I guess now we need an endpoint to get all deleted buckets (the current user can see)

DraganMitrasinovicHtec (approved):
Looks great, we just need to add permission checks, but that is a separate task

PR #141, opened 2022-07-01
SJQ2T-262 BE - Test Fixing for Edit Bucket

MarkoRadunovicHtec:
Some Test clases fixed, Some methods that are needed added Test Util classes

### DIFF
--- a/backend/src/test/java/com/htecgroup/skynest/utils/BucketEntityUtil.java
+++ b/backend/src/test/java/com/htecgroup/skynest/utils/BucketEntityUtil.java
@@ -6,6 +6,7 @@ import com.htecgroup.skynest.model.entity.UserEntity;
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
 
+import java.time.LocalDateTime;
 import java.util.UUID;
 
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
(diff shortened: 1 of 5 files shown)


backend/src/test/java/com/htecgroup/skynest/utils/BucketEntityUtil.java
```diff
@@ -18,4 +21,12 @@ public static BucketEntity getPrivateBucket() {
     bucketEntity.setCreatedBy(UserEntityUtil.getVerified());
     return bucketEntity;
   }
+
+  public static BucketEntity getOtherPrivateBucket() {
+    BucketEntity bucketEntity = getPrivateBucket();
+    UserEntity userEntity = UserEntityUtil.getVerified();
+    userEntity.setId(UUID.randomUUID());
```
ivanfhtec, line 29: didn't you say random uuids in tests were a problem?

MarkoRadunovicHtec: This is special case where I need to return some bucket with other owner than the loggedUser, so it's implemented in this method. Same UUID is in other used util classes so I need other to check when user is trying to edit others bucket.

DraganMitrasinovicHtec (approved):
Looks good, mostly simple fixes because of the change to the field type

PR #143, opened 2022-07-04
SJQ2T-298 BE - Create scripts for adding a user to a company

DraganMitrasinovicHtec:
Only admins can add users to companies, but there needs to be a way we can add the first admin.


ivanfhtec (approved):
works, I hope

PR #144, opened 2022-07-04
SJQ2T-309: BE - Actions can be revoked by other actions

ivanfhtec:
* added revoked_by column so that actions can be revoked by other actions
image
* also added methods to the action service


DraganMitrasinovicHtec (approved):
Love it

PR #146, opened 2022-07-04
SJQ2T-314: BE - Access setup

ivanfhtec:
* flyway, entities, and repositories for access/permission management

* shorter names for action enums

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/model/entity/AccessType.java
+++ b/backend/src/main/java/com/htecgroup/skynest/model/entity/AccessType.java
@@ -1,12 +1,14 @@
 package com.htecgroup.skynest.model.entity;
 
 import lombok.AllArgsConstructor;
+import lombok.Getter;
 
 @AllArgsConstructor
+@Getter
 public enum AccessType {
   VIEW("view"),
   DOWNLOAD("download"),
   EDIT("edit"),
   OWNER("owner");
-  public final String text;
+  final String text;
 }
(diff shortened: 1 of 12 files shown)


ivanfhtec:
image

ivanfhtec:
types of access:
* view

* download

* edit

* owner

backend/src/main/java/com/htecgroup/skynest/model/entity/AccessType.java
```diff
@@ -0,0 +1,12 @@
+package com.htecgroup.skynest.model.entity;
+
+import lombok.AllArgsConstructor;
+
+@AllArgsConstructor
+public enum AccessType {
```
petarstoimenovskihtec, line 8: Where do you use this enum?

backend/src/main/java/com/htecgroup/skynest/model/entity/AccessType.java
```diff
@@ -0,0 +1,12 @@
...
+  VIEW("view"),
+  DOWNLOAD("download"),
+  EDIT("edit"),
+  OWNER("owner");
+  public final String text;
```
petarstoimenovskihtec: probably this needs to be closed for accessing , or provide a getter if needed

petarstoimenovskihtec (approved):
With just few small comments

backend/src/main/java/com/htecgroup/skynest/model/entity/AccessType.java
ivanfhtec: nowhere at the moment, but I'm sure whoever writes the permissions service will need this enum

PR #148, opened 2022-07-05
SJQ2T-259: BE - Add Company Name to User response

DraganMitrasinovicHtec:
Made it so admins of a company can only see people in that company, and those who are not yet in any company ( so they can add them ). 
Also showing the name of a company ( or null if the user isn't in any company ) so frontend can know who can be added, and who is already working in the admins company.


DarkoVeljanoskiHtec (approved):
Looks good

PR #150, opened 2022-07-05
SJQ2T-234: BE Endpoint For Remove Folder

DarkoVeljanoskiHtec:
Added endpoint for removing folder, implementation in service as well as few tests.


backend/src/test/java/com/htecgroup/skynest/utils/FolderEntityUtil.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public static FolderEntity getFolderWithoutParent() {
+    FolderEntity folderEntity = new FolderEntity(null, bucketEntityUtil);
+    folderEntity.setName("FolderName");
+    folderEntity.setCreatedBy(UserEntityUtil.getVerified());
+    return folderEntity;
+  }
+
+  public static FolderEntity getFolderWithParent() {
+    FolderEntity folderEntity = new FolderEntity(getFolderWithoutParent(), bucketEntityUtil);
```
petarstoimenovskihtec: This creation is a good candidate for Builder pattern

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -1,22 +1,41 @@
...
-  private ModelMapper modelMapper;
   private FolderRepository folderRepository;
+  private ModelMapper modelMapper;
+
+  private ActionService actionService;
+
+  @Override
+  public void removeFolder(UUID uuid) {
+    FolderDto folderDto = modelMapper.map(folderRepository.findFolderById(uuid), FolderDto.class);
+    if (folderDto.getDeletedOn() != null) {
```
petarstoimenovskihtec: You can put 
`if (folderDto.getDeletedOn() != null)`
in a boolean method in FolderDto. It will give info if the folder is deleted or not, so the name can be isDeleted(). Maybe it can be reused in different place

backend/src/test/java/com/htecgroup/skynest/utils/FolderEntityUtil.java
```diff
@@ -5,22 +5,27 @@
...
 public final class FolderEntityUtil {
 
   protected static BucketEntity bucketEntityUtil = BucketEntityUtil.getPrivateBucket();
 
   public static FolderEntity getFolderWithoutParent() {
-    FolderEntity folderEntity = new FolderEntity(null, bucketEntityUtil);
-    folderEntity.setName("FolderName");
-    folderEntity.setCreatedBy(UserEntityUtil.getVerified());
-    return folderEntity;
+    return FolderEntity.builder().parentFolder(null).bucket(bucketEntityUtil).build();
```
petarstoimenovskihtec: parentFolder(null) is obsolete since the default value is null right?

backend/pom.xml
```diff
@@ -144,6 +144,13 @@
             <version>3.8.0</version>
             <scope>test</scope>
         </dependency>
+        <!-- https://mvnrepository.com/artifact/org.inferred/freebuilder -->
```
petarstoimenovskihtec: Why is this lib needed?

DarkoVeljanoskiHtec: I just remembered to remove this, it's totally unnecessary.

backend/src/main/java/com/htecgroup/skynest/model/dto/FolderDto.java
```diff
@@ -26,4 +24,13 @@ public class FolderDto {
   @With private FolderDto parentFolder;
 
   @With private BucketDto bucket;
+
+  public FolderDto deleteFolder() {
+    return this.withDeletedOn(LocalDateTime.now());
+  }
+
+  public boolean isDeleted() {
+    if (deletedOn != null) return true;
```
DraganMitrasinovicHtec: Why not just `return deletedOn != null;`

DarkoVeljanoskiHtec: I left it like this on purpose for you to catch it, nice job !

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -67,6 +74,21 @@ private FolderEntity createFolder(
     return folderEntity;
   }
 
+  @Override
+  public void removeFolder(UUID uuid) {
+    FolderDto folderDto =
+        modelMapper.map(
+            folderRepository.findById(uuid).orElseThrow(FolderAlreadyDeletedException::new),
```
DraganMitrasinovicHtec: Shouldn't this throw a FolderNotFoundException instead?

backend/src/main/java/com/htecgroup/skynest/exception/folder/FolderAlreadyDeletedException.java
```diff
@@ -0,0 +1,12 @@
+package com.htecgroup.skynest.exception.folder;
+
+import com.htecgroup.skynest.exception.SkyNestBaseException;
+import org.springframework.http.HttpStatus;
+
+public class FolderAlreadyDeletedException extends SkyNestBaseException {
+  public static final String MESSAGE = "Folder already deleted";
+
+  public FolderAlreadyDeletedException() {
+    super(MESSAGE, HttpStatus.NOT_FOUND);
```
DraganMitrasinovicHtec: Maybe 409 (Conflict) is better here, as FolderNotFound would throw a 404. Also I think you wrote in swagger that this is a 409.

DarkoVeljanoskiHtec: Yes I already change it in my other PR, but I forgot on this one.

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
```diff
@@ -168,4 +165,68 @@ public ResponseEntity<FolderResponse> getFolderDetails(@PathVariable UUID uuid)
...
+                  examples = {
+                    @ExampleObject(
+                        value =
+                            "{\"messages\":[\"Folder is already removed.\"],"
+                                + " \"status\": \"409\","
+                                + " \"timestamp\": \"2022-06-07 16:18:12\"}")
+                  })
+            })
+      })
+  @PutMapping("delete/{uuid}")
```
petarstoimenovskihtec, line 227: Why not @DeleteMapping? You can look this comment here and do the same thing https://github.com/JDP2022SkyNest/SkyNest/pull/166#discussion_r916586485

petarstoimenovskihtec (approved):

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
DarkoVeljanoskiHtec: because it's soft delete, we only change the value of deletedOn in the DB, so it's updating not really deleting.

PR #151, opened 2022-07-05
Fixing tests and get all users without company

DraganMitrasinovicHtec:
Fixed tests and added a separate case for when a user with no company calls getAllUsers.


MarkoRadunovicHtec (approved):
Seems okay!

PR #154, opened 2022-07-05
Fix for LoggedUserResponse

DraganMitrasinovicHtec:
This endpoint was throwing a 500 error, changed response to have the String companyName instead of CompanyDto


MarkoRadunovicHtec (approved):
Quick fix.

PR #156, opened 2022-07-05
SJQ2T-198 Endpoint for view bucket content

MarkoRadunovicHtec:


backend/src/main/java/com/htecgroup/skynest/model/response/FolderFileResponse.java
```diff
@@ -0,0 +1,14 @@
+package com.htecgroup.skynest.model.response;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+import java.util.List;
+
+@Data
+@AllArgsConstructor
+public class FolderFileResponse {
```
petarstoimenovskihtec: Why not renaming it to BucketResponse ?

petarstoimenovskihtec: or BucketContentResponse because you already have the previous one

MarkoRadunovicHtec: Yes, BucketContentResponse make more sense.. Will change it :)

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
```diff
@@ -137,4 +141,24 @@ public BucketResponse editBucket(
     actionService.recordAction(Collections.singleton(savedBucketEntity), ActionType.EDIT);
     return modelMapper.map(savedBucketEntity, BucketResponse.class);
   }
+
+  @Override
+  public FolderFileResponse getBucketContent(UUID bucketId) {
+
+    List<FolderEntity> allFolders =
```
petarstoimenovskihtec: I would move the allFolders and corresponding mapper to separate method in folderService and the same for allFiles, that way it can be reusable and the whole logic wont be in one place

MarkoRadunovicHtec: Great suggestion. Will do it. It will be later used again in FolderContentResponse so it is really better that way

ivanfhtec: something like getAllFiles(bucketId, parentFolderId)?

in the file service

backend/src/main/java/com/htecgroup/skynest/model/response/FolderFileResponse.java
ivanfhtec: it's just that getFolderContents will return the exact same response

if we call it BucketContentResponse, we can't really reuse it

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
ivanfhtec: actually, there should be two methods
getAllFilesFromRoot(bucketId)
getAllFilesFromFolder(parentFolderId)

same for folders

backend/src/main/java/com/htecgroup/skynest/model/response/FolderFileResponse.java
petarstoimenovskihtec: ok decide on the name, but `FolderFileResponse` to be honest doesn't tell me anything about the content of that object

backend/src/main/java/com/htecgroup/skynest/model/response/StorageContentResponse.java
```diff
@@ -0,0 +1,14 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+import java.util.List;
+
+@Data
+@AllArgsConstructor
+public class StorageContentResponse {
+
+  private List<FolderResponse> folderResponseList;
```
petarstoimenovskihtec: if it is list you can rename the variable with plural without the List suffix for example `folders` and `files`. It is more intuitive that way

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -101,6 +103,26 @@ public FileDownloadResponse downloadFile(UUID fileId) {
         fileMetadataEntity.getName(), fileMetadataEntity.getType(), fileContents);
   }
 
+  @Override
+  public List<FileResponse> getAllFilesFromRoot(UUID bucketId) {
```
petarstoimenovskihtec: maybe again rename it to `getAllRootFiles`

petarstoimenovskihtec: Same can be applied for every other method

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -101,6 +103,26 @@ public FileDownloadResponse downloadFile(UUID fileId) {
...
+  }
+
+  @Override
+  public List<FileResponse> getAllFilesFromParent(UUID parentFolderId) {
+    List<FileMetadataEntity> allFiles =
+        fileMetadataRepository.findAllByParentFolderId(parentFolderId);
+    return getFileResponses(allFiles);
+  }
+
+  private List<FileResponse> getFileResponses(List<FileMetadataEntity> allFiles) {
```
petarstoimenovskihtec: this one, maybe getFiles is better, right? `getFileReponses`  me personally leads me to think that there will be some response from files, I know it sounds not logicall when you put that in a sentence, but you get my point :)

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -101,6 +103,26 @@ public FileDownloadResponse downloadFile(UUID fileId) {
...
 
+  @Override
+  public List<FileResponse> getAllFilesFromRoot(UUID bucketId) {
+    List<FileMetadataEntity> allFiles =
+        fileMetadataRepository.findAllByBucketIdAndParentFolderIsNull(bucketId);
+    return getFileResponses(allFiles);
+  }
+
+  @Override
+  public List<FileResponse> getAllFilesFromParent(UUID parentFolderId) {
```
petarstoimenovskihtec: Aren't those files at same level?

backend/src/main/java/com/htecgroup/skynest/model/response/StorageContentResponse.java
MarkoRadunovicHtec: Will change it.

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
MarkoRadunovicHtec: Sure. :)

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
MarkoRadunovicHtec: I get it. Will change it. :)

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
MarkoRadunovicHtec: Yes but that level is determined by the parent folder. We are returning first level of that particular folder sent by parameter.

MarkoRadunovicHtec: I renamed it also, I wont use it in this PR but in getFolderContent later it will be used so I added it now. Same for Folders.

backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
```diff
@@ -164,9 +165,9 @@ public ResponseEntity<BucketResponse> createBucket(
                   examples = {@ExampleObject(value = "Internal Server Error")})
             })
       })
-  @GetMapping("/{uuid}")
+  @GetMapping("/{uuid}/info")
   public ResponseEntity<BucketResponse> getBucket(@PathVariable UUID uuid) {
-    BucketResponse bucketResponse = bucketService.getBucket(uuid);
+    BucketResponse bucketResponse = bucketService.getBucketDetails(uuid);
```
MetoMladenovski: You can change it to bucketId, since we are changing every endpoint.

MarkoRadunovicHtec: Of course.

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -74,4 +78,23 @@ public FolderResponse getFolderDetails(UUID uuid) {
...
+    return getFolders(allFolders);
+  }
+
+  @Override
+  public List<FolderResponse> getAllFoldersWithParent(UUID parentFolderId) {
+    List<FolderEntity> allFolders = folderRepository.findAllByParentFolderId(parentFolderId);
+    return getFolders(allFolders);
+  }
+
+  private List<FolderResponse> getFolders(List<FolderEntity> allFolders) {
```
DraganMitrasinovicHtec: I think it makes more sense for this to be named toResponse or something like that, as the function just converts a list of entities to a list of responses.

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -101,6 +103,26 @@ public FileDownloadResponse downloadFile(UUID fileId) {
...
+  }
+
+  @Override
+  public List<FileResponse> getAllFilesWithParent(UUID parentFolderId) {
+    List<FileMetadataEntity> allFiles =
+        fileMetadataRepository.findAllByParentFolderId(parentFolderId);
+    return getFiles(allFiles);
+  }
+
+  private List<FileResponse> getFiles(List<FileMetadataEntity> allFiles) {
```
DraganMitrasinovicHtec: Same as in the other comment, I find the naming confusing.

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
MarkoRadunovicHtec: Maybe toResponses or toResponseList, I agree the name maybe can decieve at first glance..

petarstoimenovskihtec (approved):

PR #161, opened 2022-07-07
SJQ2T-338: BE - Added tests for CompanyService

DraganMitrasinovicHtec:
Really should have done this when I made the service.


backend/src/test/java/com/htecgroup/skynest/service/impl/CompanyServiceImplTest.java
```diff
@@ -0,0 +1,160 @@
...
+  }
+
+  @Test
+  void addCompany_NonExistingTier() {
+    when(tierRepository.findByName(anyString())).thenReturn(Optional.ofNullable(null));
+
+    CompanyAddRequest companyAddRequest = CompanyAddRequestUtil.get();
+
+    Assertions.assertThrows(
+        NonExistingTierException.class, () -> companyService.addCompany(companyAddRequest));
```
MetoMladenovski: You can maybe assert the error message as well, since it's pretty simple and it's in the other tests.

backend/src/test/java/com/htecgroup/skynest/utils/company/CompanyEntityUtil.java
```diff
@@ -4,8 +4,22 @@
 
 public class CompanyEntityUtil extends CompanyBasicUtil {
 
-  public static CompanyEntity getCompanyEntity() {
+  public static CompanyEntity get() {
     return new CompanyEntity(
         id, createdOn, modifiedOn, deletedOn, pib, name, address, phoneNumber, email, tierEntity);
```
MarkoRadunovicHtec, line 9: Try to make some concrete id in this class like in TierEntityUtil or in class it is extending so when we need to check in other tests if it is the correct company we can reuse this method. If you know what I mean. I had some problems in some test, so I needed to refactor some util classes so I can use existing methods for easier testing.

backend/src/test/java/com/htecgroup/skynest/service/impl/CompanyServiceImplTest.java
DraganMitrasinovicHtec: Is that really needed? I mean now we are using a separate class for every exception, so the message will always be the same.

backend/src/test/java/com/htecgroup/skynest/utils/company/CompanyEntityUtil.java
DraganMitrasinovicHtec: Sure, I will make id from CompanyBasicUtil have a concrete value

backend/src/test/java/com/htecgroup/skynest/service/impl/CompanyServiceImplTest.java
petarstoimenovskihtec: It is useful to avoid that somebody didn't change the message by mistake

backend/src/test/java/com/htecgroup/skynest/service/impl/CompanyServiceImplTest.java
```diff
@@ -0,0 +1,174 @@
...
+  void addCompany_NonExistingTier() {
+    when(tierRepository.findByName(anyString())).thenReturn(Optional.empty());
+
+    CompanyAddRequest companyAddRequest = CompanyAddRequestUtil.get();
+
+    String expectedErrorMessage = NonExistingTierException.MESSAGE;
+    Exception thrownException =
+        Assertions.assertThrows(
+            NonExistingTierException.class, () -> companyService.addCompany(companyAddRequest));
+    Assertions.assertEquals(expectedErrorMessage, thrownException.getMessage());
```
ivanfhtec, line 67: you're testing whether a final field has a specific value?

backend/src/test/java/com/htecgroup/skynest/service/impl/CompanyServiceImplTest.java
```diff
@@ -0,0 +1,174 @@
...
+    when(currentUserService.getCompanyEntityFromLoggedUser()).thenReturn(Optional.empty());
+
+    CompanyEditRequest companyEditRequest1 = CompanyEditRequestUtil.get();
+
+    String expectedErrorMessage = UserNotInAnyCompanyException.MESSAGE;
+    Exception thrownException =
+        Assertions.assertThrows(
+            UserNotInAnyCompanyException.class,
+            () -> companyService.editCompany(companyEditRequest));
+    Assertions.assertEquals(expectedErrorMessage, thrownException.getMessage());
```
ivanfhtec, line 98: same question

backend/src/test/java/com/htecgroup/skynest/service/impl/CompanyServiceImplTest.java
```diff
@@ -0,0 +1,174 @@
...
+
+  @Test
+  void getMyCompany_UserNotInAnyCompany() {
+    when(currentUserService.getCompanyEntityFromLoggedUser()).thenReturn(Optional.empty());
+
+    String expectedErrorMessage = UserNotInAnyCompanyException.MESSAGE;
+    Exception thrownException =
+        Assertions.assertThrows(
+            UserNotInAnyCompanyException.class, () -> companyService.getMyCompany());
+    Assertions.assertEquals(expectedErrorMessage, thrownException.getMessage());
```
ivanfhtec, line 145: another one

backend/src/test/java/com/htecgroup/skynest/service/impl/CompanyServiceImplTest.java
```diff
@@ -0,0 +1,174 @@
...
+  void findCompanyById_CompanyNotFound() {
+    UUID uuid = CompanyEntityUtil.get().getId();
+
+    when(companyRepository.findById(any())).thenReturn(Optional.empty());
+
+    String expectedErrorMessage = CompanyNotFoundException.MESSAGE;
+    Exception thrownException =
+        Assertions.assertThrows(
+            CompanyNotFoundException.class, () -> companyService.findById(uuid));
+    Assertions.assertEquals(expectedErrorMessage, thrownException.getMessage());
```
ivanfhtec, line 122: and here

DarkoVeljanoskiHtec (approved):
good tests

PR #162, opened 2022-07-07
SJQ2T-292: BE Endpoint For View All Deleted Buckets

DarkoVeljanoskiHtec:
added endpoint, service implementation as well as query in the repo, also added juint test.


backend/src/main/java/com/htecgroup/skynest/repository/BucketRepository.java
```diff
@@ -15,4 +16,7 @@ public interface BucketRepository extends CrudRepository<BucketEntity, UUID> {
 
   @Query("SELECT SUM(b.size) FROM bucket b WHERE b.company IS NULL AND b.createdBy.id = ?1")
   Long sumSizeByUserId(UUID userId);
+
+  @Query("SELECT b FROM bucket b WHERE b.deletedOn IS NOT NULL")
+  List<BucketEntity> findAllDeleted();
```
ivanfhtec: I think findAllByDeletedIsNotNull should also work

DarkoVeljanoskiHtec: That is such random name

ivanfhtec: it's absolute filth, but then you wouldn't need to use <code>@Query</code>

ivanfhtec: your first mistake was looking at my code on the line above

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
```diff
@@ -83,6 +83,15 @@ public BucketDto findBucketByName(String name) {
     return modelMapper.map(bucketEntity, BucketDto.class);
   }
 
+  @Override
+  public List<BucketResponse> listAllDeletedBuckets() {
+    List<BucketEntity> entityList = bucketRepository.findAllByDeletedOnIsNotNull();
```
petarstoimenovskihtec, line 91: You can make` bucketRepository.findAllByDeletedOnIsNotNull()` return Set instead of List, JPA allows that by default , so you won't have conversion in next line

DarkoVeljanoskiHtec: action service in the next line is requiring set of object entity so this is complicating things as well.

backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
```diff
@@ -511,4 +511,62 @@ public ResponseEntity<StorageContentResponse> getBucketContent(@PathVariable UUI
...
+            content = {
+              @Content(
+                  mediaType = "application/json",
+                  schema = @Schema(implementation = BucketResponse.class),
+                  examples = {
+                    @ExampleObject(
+                        value =
+                            "[{\"bucketId\": \"ff52209c-f913-11ec-b939-0242ac120002\","
+                                + "\"id\": \"a6fd6d95-0a60-43ff-961f-2b9b2ff72f95\","
+                                + "\"name\": \"Name\","
```
MarkoRadunovicHtec, line 545: I think response will return createdById instead of id and also isPublic field.

DraganMitrasinovicHtec (approved):
I love it

PR #163, opened 2022-07-07
SJQ2T-316: BE - File upload/download small fixes

ivanfhtec:
* non-custom exceptions are now displayed in a bit more detail in responses

* mongo and IOStream exceptions are now logged before throwing them as custom exceptions

* fix - only bucket creators can upload/download from private buckets

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthorizationFilter.java
+++ b/backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthorizationFilter.java
@@ -22,6 +22,8 @@ import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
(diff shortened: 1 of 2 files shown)


backend/src/main/java/com/htecgroup/skynest/filter/CustomAuthorizationFilter.java
```diff
@@ -80,11 +82,20 @@ protected void doFilterInternal(
       ExceptionUtil.writeToResponse(errorMessage, response);
     } catch (Exception ex) {
       log.error(ex);
+
+      List<String> messages = new ArrayList<>();
+      Throwable currentException = ex;
+      while (currentException != null && currentException != currentException.getCause()) {
```
petarstoimenovskihtec, line 88: We will discuss this some other time, not sure if you need this logic here, but I won;t change the solution now :D

petarstoimenovskihtec (approved):

PR #164, opened 2022-07-07
SJQ2T-237: BE Endpoint For Change Folder Name 

DarkoVeljanoskiHtec:
added endpoint for edit folder name, implemented service, added test as well.

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
+++ b/backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
@@ -238,9 +238,9 @@ public class FolderController {
       })
   @PutMapping("/{uuid}")
   public ResponseEntity<FolderResponse> editFolder(
-      @Valid @RequestBody FolderEditRequest folderEditRequest, @PathVariable UUID uuid) {
+      @Valid @RequestBody FolderEditRequest folderEditRequest, @PathVariable UUID folderId) {
     ResponseEntity<FolderResponse> folderResponseEntity =
-        new ResponseEntity<>(folderService.editFolder(folderEditRequest, uuid), HttpStatus.OK);
+        new ResponseEntity<>(folderService.editFolder(folderEditRequest, folderId), HttpStatus.OK);
     return folderResponseEntity;
   }
 }
--- a/backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
+++ b/backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
@@ -27,7 +27,6 @@ import java.util.UUID;
 @Log4j2
 @Tag(name = "Folder API", description = "Folder-related operations")
 public class FolderController {
-
   private FolderService folderService;
 
   @Operation(summary = "Create new folder")
@@ -43,11 +42,9 @@ public class FolderController {
                   examples = {
                     @ExampleObject(
                         value =
-                            "{\"createdById\": \"a6fd6d95-0a60-43ff-961f-2b9b2ff72f95\","
+                            "{\"bucketId\": \"a6fd6d95-0a60-43ff-961f-2b9b2ff72f95\","
                                 + "  \"name\": \"Name\","
-                                + "  \"companyId\": \"a6fd6d95-0a60-43ff-961f-2b9b2ff72f95\","
-                                + "  \"description\": \"Description\","
-                                + "  \"size\": \"1000\"}")
+                                + "  \"parentFolderId\": \"a6fd6d95-0a60-43ff-961f-2b9b2ff72f95\"}")
                   })
             }),
         @ApiResponse(
@@ -61,7 +58,7 @@ public class FolderController {
                     @ExampleObject(
                         value =
                             "{\"messages\":[\"name cannot be null or empty\","
-                                + " \"description cannot be null or empty\"],"
+                                + " \"bucketId cannot be null or empty\"],"
                                 + " \"status\": \"400\","
                                 + " \"timestamp\": \"2022-06-07 16:18:12\"}")
                   })
@@ -139,7 +136,7 @@ public class FolderController {
             }),
         @ApiResponse(
             responseCode = "404",
-            description = "Bucket not found",
+            description = "Folder not found",
             content = {
               @Content(
                   mediaType = "application/json",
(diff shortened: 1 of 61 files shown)


backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -74,4 +80,21 @@ public FolderResponse getFolderDetails(UUID uuid) {
...
     return folderResponse;
   }
+
+  @Override
+  public FolderResponse editFolder(FolderEditRequest folderEditRequest, UUID uuid) {
+    FolderEntity folderEntity =
+        folderRepository.findById(uuid).orElseThrow(FolderNotFoundException::new);
+
+    if (folderEntity.getDeletedOn() != null) {
+      // throw new FolderAlreadyDeletedException();
```
ivanfhtec: a comment 😱

DarkoVeljanoskiHtec: I was in the process of commit and push when you saw it lol.

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
```diff
@@ -168,4 +170,78 @@ public ResponseEntity<FolderResponse> getFolderDetails(@PathVariable UUID uuid)
...
+                                + " \"status\": \"409\","
+                                + " \"timestamp\": \"2022-06-07 16:18:12\"}")
+                  })
+            })
+      })
+  @PutMapping("/{uuid}")
+  public ResponseEntity<FolderResponse> editFolder(
+      @Valid @RequestBody FolderEditRequest folderEditRequest, @PathVariable UUID uuid) {
+    ResponseEntity<FolderResponse> folderResponseEntity =
+        new ResponseEntity<>(folderService.editFolder(folderEditRequest, uuid), HttpStatus.OK);
```
MetoMladenovski: With the new changes in the endpoints, uuid should be folderId. Maybe in the service too, much more clear.

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
```diff
@@ -168,4 +170,78 @@ public ResponseEntity<FolderResponse> getFolderDetails(@PathVariable UUID uuid)
...
+  @Operation(summary = "Edit folder name")
+  @ApiResponses(
+      value = {
+        @ApiResponse(
+            responseCode = "200",
+            description = "Folder successfully edited",
+            content = {
+              @Content(
+                  mediaType = "application/json",
+                  schema = @Schema(implementation = BucketResponse.class),
```
MarkoRadunovicHtec: Isn't this FolderResponse.class?

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
```diff
@@ -168,4 +170,78 @@ public ResponseEntity<FolderResponse> getFolderDetails(@PathVariable UUID uuid)
...
+                    @ExampleObject(
+                        value =
+                            "{\"messages\":[\"Folder with id ff52209c-f913-11ec-b939-0242ac120002 doesn't exist\"],"
+                                + " \"status\": \"404\","
+                                + " \"timestamp\": \"2022-06-07 16:18:12\"}")
+                  })
+            }),
+        @ApiResponse(
+            responseCode = "409",
+            description = "Folder already deleted",
```
MetoMladenovski, line 286: You are throwing 404, but here is 409. And just as a suggestion, maybe the name should be different too, if it's not used somewhere already.

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
DarkoVeljanoskiHtec: It should be

backend/src/test/java/com/htecgroup/skynest/service/impl/FolderServiceImplTest.java
```diff
@@ -85,6 +90,35 @@ void getFolderDetails() {
...
+    when(folderRepository.findById(any())).thenReturn(Optional.of(expectedFolderEntity));
+    when(folderRepository.save(any())).thenReturn(expectedFolderEntity);
+
+    FolderEditRequest folderEditRequest = FolderEditRequestUtil.get();
+    FolderResponse actualFolderResponse =
+        folderService.editFolder(folderEditRequest, expectedFolderEntity.getId());
+
+    this.assertFolderEntityAndFolderResponse(expectedFolderEntity, actualFolderResponse);
+    verify(folderRepository, times(1)).findById(any());
+    verify(folderRepository, times(1)).save(any());
```
MetoMladenovski: I don't think you need to verify these methods since you are mocking them and you are asserting the returned entity if it has been edited. Maybe you can verify that the .recordAction method is called.

backend/src/test/java/com/htecgroup/skynest/utils/FolderEntityUtil.java
```diff
@@ -14,6 +17,7 @@ public static FolderEntity getFolderWithoutParent() {
     FolderEntity folderEntity = new FolderEntity(null, bucketEntityUtil);
     folderEntity.setName("FolderName");
     folderEntity.setCreatedBy(UserEntityUtil.getVerified());
+    folderEntity.setId(UUID.randomUUID());
```
MarkoRadunovicHtec: Try to set exact id so that tests which compares that folder id with another one have easier setup. Use UUID.fromString and add some random generated uuid value which will be always set when you call this method. We need to exclude these random id fields from test util classes for later use.

backend/src/test/java/com/htecgroup/skynest/service/impl/FolderServiceImplTest.java
```diff
@@ -85,6 +90,35 @@ void getFolderDetails() {
...
+    FolderEntity folderEntity = FolderEntityUtil.getDeletedFolder();
+    FolderEditRequest folderEditRequest = FolderEditRequestUtil.get();
+    when(folderRepository.findById(any())).thenReturn(Optional.of(folderEntity));
+    String expectedErrorMessage = FolderAlreadyDeletedException.MESSAGE;
+    Exception thrownException =
+        Assertions.assertThrows(
+            FolderAlreadyDeletedException.class,
+            () -> folderService.editFolder(folderEditRequest, UUID.randomUUID()));
+    Assertions.assertEquals(expectedErrorMessage, thrownException.getMessage());
+    verify(folderRepository, times(1)).findById(any());
```
MetoMladenovski, line 144: Same goes here.

backend/src/test/java/com/htecgroup/skynest/service/impl/FolderServiceImplTest.java
MarkoRadunovicHtec: But he verify that the methods are called just once. It is useful if method is changed so somehow some methods are called more times than needed maybe and then it is recorded here and test will fail.

MetoMladenovski: Well I'm not sure, but I think you can either use when() to mock them, or use verify() to verify them. I even think that if it gets called twice and we have it mocked in when() once, it will fail. My point is I think it's redundant to use when() and verify() on the same method.

ivanfhtec: @petarstoimenovskihtec @markobarackovhtec  we need guidance

MarkoRadunovicHtec: I tried that in some tests before I think. You only need to mock it once, every time it's called it will return mocked type. But verify checks how many times it was called. We will get guidance tomorrow I guess. :D

petarstoimenovskihtec: Well _when_  and _verify_ are two completely different methods with different usage. 
You can use `when` to define what is going to be returned when that specific method is called. 
You can mock it, to return the same value no matter how many times you call it and which input params you provide with `any()` or you can mock it to return specific value for different input param.
Example:
`when(folderRepository.findById(any())).thenReturn(MockObject1);`
and you call the method with
```
 folderRepository.findById(id1);  
 folderRepository.findById(id2); 
```
the result will be in both cases `MockObject1`.
The other option is:
```
when(folderRepository.findById(id1)).thenReturn(MockObject1);
when(folderRepository.findById(id2)).thenReturn(MockObject2);
```
If you have mocked responses like this, then if you have the same situation as above you will get the different responses aka _MockObject1_ and _MockObject2_.

You can further more defines mock to return one object first time and different object second time when it is called, no matter the input param (it is useful if you need to test recursion or something that has loop ).
`when(folderRepository.findById(any())).thenReturn(MockObject1).thenReturn(OtherMockObject2);`

On the other hand, _verify_  is used to do the verification that the method is called that many times as you expect with the corresponding value. Meaning in this case the verify method instead of any() should have the exact value if possible. That way you can be sure that the method was actually called  with the parameters that you expected.

I hope I made this a little bit clearer :)

MetoMladenovski: Definitely clearer. My mind was super focused on this kind of a situation that i totally forgot about any()..
```
when(folderRepository.findById(id1)).thenReturn(MockObject1);
when(folderRepository.findById(id2)).thenReturn(MockObject2);
```
So it is a good practice to verify every method?

backend/src/test/java/com/htecgroup/skynest/utils/FolderEntityUtil.java
DarkoVeljanoskiHtec: Actually id is not needed here at all, I was trying something and forgot to remove it, I will take it out, but when we need id in the future we'll do it like you say.

backend/src/test/java/com/htecgroup/skynest/service/impl/FolderServiceImplTest.java
petarstoimenovskihtec: Well yeah, it is always good to verify anything, makes more sense if you are doing it with expected value instead of any(). When using when() you can use any() if you don't have specific behaviour for some input, and  with verify() I strongly suggest to use the expected value. Also it is always useful to verify that there aren't any additional interactions with that mock.
` verifyNoMoreInteractions(mockServiceHere);`

ivanfhtec (approved):

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -100,6 +101,23 @@ public FolderResponse getFolderDetails(UUID uuid) {
     return folderResponse;
   }
 
+  @Override
+  public FolderResponse editFolder(FolderEditRequest folderEditRequest, UUID folderId) {
+    FolderEntity folderEntity =
+        folderRepository.findById(folderId).orElseThrow(FolderNotFoundException::new);
+
+    if (folderEntity.getDeletedOn() != null) {
```
DraganMitrasinovicHtec: Would be nice if you make a method for this. I like isDeleted. I think we did it that way before

DarkoVeljanoskiHtec: see now if you like it better

PR #165, opened 2022-07-07
SJQ2T-337: Clearer endpoint names

ivanfhtec:
* switched to bucketId, folderId, fileId… instead of uuid

* ids go before operations, so <code>users/{userId}/password-change/</code> rather than <code>users/password-change/{userId}</code>

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
+++ b/backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
@@ -103,7 +103,7 @@ public class BucketController {
     return bucketResponseEntity;
   }
 
-  @Operation(summary = "Get bucket with given uuid")
+  @Operation(summary = "Get bucket details")
   @ApiResponses(
       value = {
         @ApiResponse(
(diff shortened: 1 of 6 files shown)


ivanfhtec:
image
image
image
image

petarstoimenovskihtec:
This is something that has to be done at the same time with QA and FE teams. Because if you merge this PR they will have issues in there parts especially in User related endpoints

ivanfhtec:
> This is something that has to be done at the same time with QA and FE teams. Because if you merge this PR they will have issues in there parts especially in User related endpoints

Most of these changes only affect the documentation. But there are a couple that FE and QA have to change as well
* GET /buckets/{uuid} → /buckets/{uuid}/info

* DELETE /buckets/delete/{uuid} → /buckets/{uuid}/delete

* PUT /users/password-change/{uuid} → /users/{uuid}/password-change

DarkoVeljanoskiHtec (approved):
looks better

backend/src/main/java/com/htecgroup/skynest/controller/BucketController.java
```diff
@@ -288,13 +288,13 @@ public List<BucketResponse> getBuckets() {
                   })
             })
       })
-  @PutMapping("delete/{uuid}")
-  public ResponseEntity<Boolean> deleteBucket(@PathVariable UUID uuid) {
-    bucketService.deleteBucket(uuid);
+  @PutMapping("/{bucketId}/delete")
```
petarstoimenovskihtec, line 291: Since you are changing this, I would change this to be Delete http, for consisted level around the app

PR #166, opened 2022-07-08
SJQ2T-247: BE -  Endpoint for file removal

DraganMitrasinovicHtec:
Endpoint for deleting files, permission checks are to be added later.


backend/src/main/java/com/htecgroup/skynest/controller/FileController.java
```diff
@@ -366,4 +366,95 @@ public ResponseEntity<FileResponse> getFileDetails(@PathVariable(name = "uuid")
...
+                  mediaType = MediaType.APPLICATION_JSON_VALUE,
+                  schema = @Schema(implementation = String.class),
+                  examples = {@ExampleObject(value = "Internal Server Error")})
+            })
+      })
+  @DeleteMapping("/{fileId}")
+  public ResponseEntity<Boolean> deleteFile(@PathVariable UUID fileId) {
+
+    FileResponse fileResponse = fileService.deleteFile(fileId);
+    log.info("File {} successfully deleted", fileResponse.getName());
```
ivanfhtec: record this as a delete action in the service method

backend/src/test/java/com/htecgroup/skynest/service/impl/FileServiceImplTest.java
```diff
@@ -60,4 +62,27 @@ void downloadFile_ThrowsFileNotFound() {
...
+    when(fileMetadataRepository.findById(fileId)).thenReturn(Optional.empty());
+
+    Assertions.assertThrows(FileNotFoundException.class, () -> fileService.deleteFile(fileId));
+    verify(fileMetadataRepository, times(1)).findById(any());
+  }
+
+  @Test
+  void deleteFile_ThrowsFileAlreadyDeleted() {
+    FileMetadataEntity fileMetadataEntity = FileMetadataEntityUtil.getFileMetadataEntity();
+    fileMetadataEntity.setDeletedOn(LocalDateTime.now());
```
ivanfhtec: you could probably make a util method that returns a deleted file

backend/src/test/java/com/htecgroup/skynest/service/impl/FileServiceImplTest.java
```diff
@@ -60,4 +62,27 @@ void downloadFile_ThrowsFileNotFound() {
...
   }
+
+  @Test
+  void deleteFile_ThrowsFileNotFound() {
+    UUID fileId = FileMetadataEntityUtil.getFileMetadataEntity().getId();
+
+    when(fileMetadataRepository.findById(fileId)).thenReturn(Optional.empty());
+
+    Assertions.assertThrows(FileNotFoundException.class, () -> fileService.deleteFile(fileId));
+    verify(fileMetadataRepository, times(1)).findById(any());
```
ivanfhtec: I guess you could also verify the action wasn't recorded

backend/src/main/java/com/htecgroup/skynest/controller/FileController.java
```diff
@@ -366,4 +366,95 @@ public ResponseEntity<FileResponse> getFileDetails(@PathVariable(name = "uuid")
...
+            description = "Internal Server Error",
+            content = {
+              @Content(
+                  mediaType = MediaType.APPLICATION_JSON_VALUE,
+                  schema = @Schema(implementation = String.class),
+                  examples = {@ExampleObject(value = "Internal Server Error")})
+            })
+      })
+  @DeleteMapping("/{fileId}")
+  public ResponseEntity<Boolean> deleteFile(@PathVariable UUID fileId) {
```
petarstoimenovskihtec: It not that something is wrong but instead of returning true as body, you can return status code 204 . It indicates that the server has successfully fulfilled the request and that there is no content to send in the response payload body.
You can use an Annotation for that as well :)

backend/src/main/java/com/htecgroup/skynest/model/entity/ObjectEntity.java
```diff
@@ -33,4 +33,8 @@ public class ObjectEntity implements Serializable {
   private UserEntity createdBy;
 
   private String name;
+
+  public boolean isDeleted() {
```
petarstoimenovskihtec, line 37: 👍

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -101,6 +103,19 @@ public FileDownloadResponse downloadFile(UUID fileId) {
...
   }
 
+  @Override
+  public FileResponse deleteFile(UUID fileId) {
+
+    FileMetadataEntity fileMetadataEntity = getFileMetadataEntity(fileId);
+    if (fileMetadataEntity.isDeleted()) {
+      throw new FileAlreadyDeletedException();
+    }
+    fileMetadataEntity.setDeletedOn(LocalDateTime.now());
```
petarstoimenovskihtec: You can do the same as isDeleted() here :)

backend/src/test/java/com/htecgroup/skynest/service/impl/FileServiceImplTest.java
petarstoimenovskihtec: Since you already mocked this method to return response only for a specific input, why not use the same input here as well ? to verify that indeed it was called with that value.

Just for test purposes to see how it works, you can add some random value in verify method and see what will happen with the test execution

backend/src/main/java/com/htecgroup/skynest/controller/FileController.java
DraganMitrasinovicHtec: Totally forgot, thanks for noticing

petarstoimenovskihtec: I have one a little bit advanced way for you to do it.. 
You can create a Custom annotation service layer for delete action and provide the specific input params there.
That custom annotation would be handled by AOP (Aspect oriented programming - module provided by Spring) and that Aspect would be triggered after the result of the service is executed. 

Don't have to do it now, but investigate it and I believe you have never used that approach and you will find it interesting and maybe it will give you different perspectives :)

backend/src/test/java/com/htecgroup/skynest/service/impl/FileServiceImplTest.java
DraganMitrasinovicHtec: I somehow forgot what I was doing 2 lines ago :D. Definitely no need to use any here

backend/src/main/java/com/htecgroup/skynest/controller/FileController.java
DraganMitrasinovicHtec: Yeah, we didn't really know what's the best practice with delete functions. Is it alright for it to be a void function with @ResponseStatus(HttpStatus.NO_CONTENT)?

petarstoimenovskihtec: yup

backend/src/main/java/com/htecgroup/skynest/service/impl/ActionServiceImpl.java
```diff
@@ -106,4 +120,21 @@ public List<ActionEntity> getActionsWithTypeForObject(
       return actionRepository.findAllByObjectIdAndTypeAndNotRevokedOrderByPerformedOnDesc(
           objectId, actionTypeEntity);
   }
+
+  @AfterReturning("@annotation(recordAction)")
```
DraganMitrasinovicHtec: @petarstoimenovskihtec Is this what you had in mind? I add the RecordAction annotation on every method which is supposed to record one, and from the annotation paramaters I get the Action type and object id.

petarstoimenovskihtec: yup, something like that

petarstoimenovskihtec: You only need to move that aop logic into separate component, not here in this service. You can create new package named aspects and move this logic in some new class there, maybe ActionsAop or something like that

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -126,6 +128,19 @@ private List<FileResponse> asFileResponseList(List<FileMetadataEntity> allFiles)
         .collect(Collectors.toList());
   }
 
+  @Override
+  @RecordAction(objectId = "[0].toString()", actionType = ActionType.DELETE)
+  public FileResponse deleteFile(UUID fileId) {
```
petarstoimenovskihtec: Does every method that needs to trigger this annotation has the same input params?

backend/src/main/java/com/htecgroup/skynest/service/impl/ActionServiceImpl.java
DraganMitrasinovicHtec: Makes sense

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
DraganMitrasinovicHtec: No, I intend to use this for every sort of action ( edits, downloads ...). That's why I need to send the actionType

backend/src/main/java/com/htecgroup/skynest/controller/FileController.java
```diff
@@ -366,4 +366,94 @@ public ResponseEntity<FileResponse> getFileDetails(@PathVariable UUID fileId) {
...
+                  schema = @Schema(implementation = String.class),
+                  examples = {@ExampleObject(value = "Internal Server Error")})
+            })
+      })
+  @DeleteMapping("/{fileId}")
+  @ResponseStatus(HttpStatus.NO_CONTENT)
+  public void deleteFile(@PathVariable UUID fileId) {
+
+    FileResponse fileResponse = fileService.deleteFile(fileId);
+    log.info("File {} successfully deleted", fileResponse.getName());
```
ivanfhtec: remove since you're using actions now

ivanfhtec (approved):

PR #168, opened 2022-07-08
SJQ2T-312 Endpoint for View Folder Content

MarkoRadunovicHtec:
Initial endpoint created with swagger documentation and service layer. Implemented the rest. Tests added and util classes modified for the test needs.


backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -113,6 +117,15 @@ public List<FolderResponse> getAllFoldersWithParent(UUID parentFolderId) {
     return asFolderResponseList(allFolders);
   }
 
+  @Override
+  public StorageContentResponse getFolderContent(UUID folderId) {
```
DraganMitrasinovicHtec, line 140: You should check if the folder id exists, this way it returns an empty list for an invalid folder id. Some exception should be thrown

MarkoRadunovicHtec: Makes sense.. This way you don't know if the folder exists at all or it is just empty.

DraganMitrasinovicHtec (approved):
Looks good to me

PR #171, opened 2022-07-08
SJQ2T-240: BE - Upload file to folder

ivanfhtec:


backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -126,29 +145,60 @@ private List<FileResponse> asFileResponseList(List<FileMetadataEntity> allFiles)
...
+    emptyFileMetadata.setBucket(bucketEntity);
+    emptyFileMetadata.setParentFolder(null);
+
+    return emptyFileMetadata;
+  }
+
+  private FileMetadataEntity initFileMetadataWithFolder(
+      MultipartFile multipartFile, UUID folderId) {
+
+    FileMetadataEntity emptyFileMetadata =
```
petarstoimenovskihtec: Since this 2 methods use pretty much the same logic (initFileMetadataWithBucket and initFileMetadataWithFolder ), I would extract the mutual logic into a new private method ,maybe named i_nitFileMetadata(MultipartFile, BucketId,  FolderId)_ and in this separate method just call that method with the corresponding input params

ivanfhtec: so, just pass a null for either parameter to this private method, which should be enough for it to decide what to do

I started off with something like that, but it felt a bit confusing having to pass a null

ivanfhtec: I did some refactoring

does everything else look good?

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -126,29 +155,49 @@ private List<FileResponse> asFileResponseList(List<FileMetadataEntity> allFiles)
...
+      bucket = parentFolder.getBucket();
+    }
+    fileMetadataEntity.setBucket(bucket);
+    fileMetadataEntity.setParentFolder(parentFolder);
+
     return fileMetadataEntity;
   }
 
+  private FileMetadataEntity initFileMetadataWithBucket(
+      String name, long size, String type, UUID bucketId) {
```
petarstoimenovskihtec, line 215: Why not put this  `bucketRepository.findById(bucketId).orElseThrow(BucketNotFoundException::new);`
here and the same corresponding one in the other method
Also this two lines will be moved into _**initFileMetadata**_
```
 fileMetadataEntity.setBucket(bucket);
    fileMetadataEntity.setParentFolder(parentFolder);
```

ivanfhtec: makes sense

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -66,6 +70,31 @@ public FileResponse uploadFile(MultipartFile multipartFile, UUID bucketId) {
...
+  public FileResponse uploadFileToFolder(MultipartFile multipartFile, UUID folderId) {
+
+    FileMetadataEntity emptyFileMetadata =
+        initFileMetadataWithFolder(
+            multipartFile.getOriginalFilename(),
+            multipartFile.getSize(),
+            multipartFile.getContentType(),
+            folderId);
+
+    checkOnlyCreatorsCanAccessPrivateBuckets(emptyFileMetadata);
```
petarstoimenovskihtec, line 92: Just a note:

I would consider when refactoring this class to explore the option for something like emptyFileMetadata.validate(). Or maybe to extract this in some class FileMetadataDto or however you are naming the models for shifting data between repository and controller layers and make a validate method there

petarstoimenovskihtec (approved):

PR #177, opened 2022-07-11
SJQ2T-304: BE - Grant bucket permissions

ivanfhtec:
* users can now grant permissions over a bucket, if they are owners
* removed sql dialect from application.properties since the default (mysql8) works fine
  *  and we're actually using mysql 8, not 5
* sort of fixed tests

I'll write tests if this seems good

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/exception/accesstype/AccessTypeException.java
+++ b/backend/src/main/java/com/htecgroup/skynest/exception/accesstype/AccessTypeException.java
@@ -1,0 +1,13 @@
+package com.htecgroup.skynest.exception.accesstype;
+
+import com.htecgroup.skynest.exception.SkyNestBaseException;
+import org.springframework.http.HttpStatus;
+
+public class AccessTypeException extends SkyNestBaseException {
+
+  private static final long serialVersionUID = -6920883293393177324L;
+
+  protected AccessTypeException(String message, HttpStatus status) {
+    super(message, status);
+  }
+}
--- a/backend/src/main/java/com/htecgroup/skynest/model/entity/UserObjectAccessKey.java
+++ b/backend/src/main/java/com/htecgroup/skynest/model/entity/UserObjectAccessKey.java
@@ -4,10 +4,12 @@ import lombok.AllArgsConstructor;
 import lombok.EqualsAndHashCode;
 import lombok.NoArgsConstructor;
 
-import javax.persistence.JoinColumn;
-import javax.persistence.ManyToOne;
+import javax.persistence.Column;
+import javax.persistence.Embeddable;
 import java.io.Serializable;
+import java.util.UUID;
 
+@Embeddable
 @NoArgsConstructor
 @AllArgsConstructor
 @EqualsAndHashCode
@@ -15,11 +17,9 @@ public class UserObjectAccessKey implements Serializable {
 
   private static final long serialVersionUID = -1295532429399577410L;
 
-  @ManyToOne
-  @JoinColumn(name = "granted_to", nullable = false)
-  private UserEntity grantedTo;
+  @Column(name = "granted_to")
+  private UUID grantedToId;
 
-  @ManyToOne
-  @JoinColumn(name = "object_id", nullable = false)
-  private ObjectEntity object;
+  @Column(name = "object_id")
+  private UUID objectId;
 }
--- a/backend/src/main/java/com/htecgroup/skynest/model/request/PermissionGrantRequest.java
+++ b/backend/src/main/java/com/htecgroup/skynest/model/request/PermissionGrantRequest.java
@@ -1,0 +1,24 @@
+package com.htecgroup.skynest.model.request;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
+import java.util.UUID;
+
+@Data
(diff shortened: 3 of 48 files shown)


backend/src/main/java/com/htecgroup/skynest/model/request/PermissionGrantRequest.java
```diff
@@ -0,0 +1,24 @@
...
+@Data
+@NoArgsConstructor
+@AllArgsConstructor
+public class PermissionGrantRequest {
+
+  @NotNull(message = "user id cannot be null or empty")
+  private UUID grantedTo;
+
+  @NotNull(message = "bucket id cannot be null or empty")
+  private UUID bucketId;
```
DraganMitrasinovicHtec: Good enought for this PR, but later we can maybe add a generic way for granting permissions.

backend/src/main/java/com/htecgroup/skynest/model/response/PermissionResponse.java
```diff
@@ -0,0 +1,25 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+
+import java.util.UUID;
+
+@Data
+@AllArgsConstructor
+@NoArgsConstructor
+public class PermissionResponse {
```
DraganMitrasinovicHtec, line 12: Do we need both Id and email here?

ivanfhtec: should prove useful to FE

backend/src/main/java/com/htecgroup/skynest/model/request/PermissionGrantRequest.java
ivanfhtec: this really should be objectId, for future (re)use

ivanfhtec: it's changed

backend/src/main/java/com/htecgroup/skynest/exception/accesstype/AccessTypeException.java
```diff
@@ -0,0 +1,13 @@
+package com.htecgroup.skynest.exception.accesstype;
+
+import com.htecgroup.skynest.exception.SkyNestBaseException;
+import org.springframework.http.HttpStatus;
+
+public class AccessTypeException extends SkyNestBaseException {
+
+  private static final long serialVersionUID = -6920883293393177324L;
```
DarkoVeljanoskiHtec, line 8: Whats the deal with all this serial ids inside all exceptions ?

ivanfhtec: Throwable is Serializable

our exceptions extend Throwable 😔

DarkoVeljanoskiHtec: We didn't have one in the previous exceptions, thats why I'm asking.

backend/src/main/java/com/htecgroup/skynest/model/entity/UserObjectAccessKey.java
```diff
@@ -4,22 +4,22 @@
...
 import lombok.NoArgsConstructor;
 
-import javax.persistence.JoinColumn;
-import javax.persistence.ManyToOne;
+import javax.persistence.Column;
+import javax.persistence.Embeddable;
 import java.io.Serializable;
+import java.util.UUID;
 
+@Embeddable
```
DarkoVeljanoskiHtec, line 12: How are this two entities connected, this one is embeddable and the other one has Embedded id from this one, how that works?

ivanfhtec: `UserObjectAccessEntity` has a composite primary key

when your table has a composite primary key, you have use embeddedIds (though there are other ways to do this)
I've made `UserObjectAccessKey` `@Embeddable`, and it serves as the `@EmbeddedId` in `UserObjectAccessEntity`

ivanfhtec: unfortunately, this composite key happens to be made up of two foreign keys
you can use <code>@MapsId</code> to map these foreign keys to field in <code>@EmbeddedId</code>

ivanfhtec: we need this because the primary key is composite
```
@EmbeddedId private UserObjectAccessKey id;
```

we need this because these are foreign keys
```
@MapsId("grantedToId")
@ManyToOne
@JoinColumn(name = "granted_to", nullable = false)
private UserEntity grantedTo;

@MapsId("objectId")
@ManyToOne
@JoinColumn(name = "object_id", nullable = false)
private ObjectEntity object;
```

ivanfhtec: `@MapsId("grantedToId")` maps this foreign key to the composite key, where `grantedToId` is a field in `UserObjectAccessKey`

DarkoVeljanoskiHtec: I understand now, thanks.

DarkoVeljanoskiHtec (approved):
Looks good enough 👍

PR #182, opened 2022-07-12
SJQ2T-256: BE Endpoint For Move Folder

DarkoVeljanoskiHtec:
two endpoints for move folder, one for moving in into bucket, one for moving in into bucket, the needed services are also done, as well as tests. (I need to add and change some tests)


backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
```diff
@@ -233,6 +235,136 @@ public void removeFolder(@PathVariable UUID uuid) {
...
+                  examples = {
+                    @ExampleObject(
+                        value =
+                            "{\"messages\":[\"Folder is already inside the bucket.\"],"
+                                + " \"status\": \"409\","
+                                + " \"timestamp\": \"2022-06-07 16:18:12\"}")
+                  })
+            })
+      })
+  @PutMapping("move/{uuid}/toBucket")
```
ivanfhtec: I'm not sure we even need to move stuff between buckets

also, right now the database won't adjust the bucket's size if you move something to a different bucket
you'd need to change the triggers

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
```diff
@@ -233,6 +235,136 @@ public void removeFolder(@PathVariable UUID uuid) {
...
+                  examples = {
+                    @ExampleObject(
+                        value =
+                            "{\"messages\":[\"Folder is already inside the folder.\"],"
+                                + " \"status\": \"409\","
+                                + " \"timestamp\": \"2022-06-07 16:18:12\"}")
+                  })
+            })
+      })
+  @PutMapping("move/{uuid}/toFolder")
```
ivanfhtec: why not `/{folderId}/move/{destinationFolderId}` or `/{folderId}/move/destination/{destinationFolderId}`

where destinationFolderId could be null as well

ivanfhtec (commented):
initial feedback

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
ivanfhtec: or even `/{folderId}/move/folders/{destinationFolderId}`, so the whole thing is `/folders/{folderId}/move/folders/{destinationFolderId}`

DarkoVeljanoskiHtec: why destinationFolderId would be null ?

ivanfhtec: if you're moving a folder to the "root" folder - the bucket

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
DarkoVeljanoskiHtec: I agree we should drop the move folder to bucket functionality.

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -136,6 +140,38 @@ public List<FolderResponse> getAllFoldersWithParent(UUID parentFolderId) {
...
+    BucketEntity bucketEntity =
+        bucketRepository
+            .findById(moveFolderToBucketRequest.getDestinationBucketId())
+            .orElseThrow(BucketNotFoundException::new);
+    if (folderEntity.getBucket().getId() == bucketEntity.getId()) {
+      throw new FolderAlreadyInsideBucketException();
+    }
+    folderEntity.setBucket(bucketEntity);
+    folderRepository.save(folderEntity);
+    actionService.recordAction(Collections.singleton(folderEntity), ActionType.MOVE);
```
petarstoimenovskihtec: Wasn't there an annotation that does this action logic?

DarkoVeljanoskiHtec: Yes, but it's on different PR, still not on dev.

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
```diff
@@ -233,6 +233,135 @@ public void removeFolder(@PathVariable UUID folderId) {
     folderService.removeFolder(folderId);
   }
 
+  @Operation(summary = "Move Folder")
```
ivanfhtec: this is move folder to root?

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
```diff
@@ -233,6 +233,135 @@ public void removeFolder(@PathVariable UUID folderId) {
...
+                  })
+            })
+      })
+  @PutMapping("/{folderId}/move")
+  @ResponseStatus(HttpStatus.NO_CONTENT)
+  public void moveFolderToRoot(@PathVariable UUID folderId) {
+    folderService.moveFolderToBucket(folderId);
+  }
+
+  @Operation(summary = "Move Folder")
```
ivanfhtec: move folder to folder?

backend/src/main/java/com/htecgroup/skynest/exception/folder/FolderAlreadyInsideBucketException.java
```diff
@@ -0,0 +1,13 @@
+package com.htecgroup.skynest.exception.folder;
+
+import com.htecgroup.skynest.exception.SkyNestBaseException;
+import org.springframework.http.HttpStatus;
+
+public class FolderAlreadyInsideBucketException extends SkyNestBaseException {
```
ivanfhtec: do we need this?

backend/src/main/java/com/htecgroup/skynest/service/FolderService.java
```diff
@@ -24,5 +24,9 @@ FolderResponse createFolder(
 
   List<FolderResponse> getAllFoldersWithParent(UUID parentFolderId);
 
+  void moveFolderToBucket(UUID folderId);
```
ivanfhtec: move folder to bucket root?

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -136,6 +138,33 @@ public List<FolderResponse> getAllFoldersWithParent(UUID parentFolderId) {
     return asFolderResponseList(allFolders);
   }
 
+  @Override
+  public void moveFolderToBucket(UUID folderId) {
+    FolderEntity folderEntity =
+        folderRepository.findById(folderId).orElseThrow(FolderNotFoundException::new);
+    if (folderEntity.getParentFolder() == null) {
+      throw new FolderAlreadyInsideBucketException();
```
ivanfhtec: why not throw FolderAlreadyInsideFolder here?

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -136,6 +138,33 @@ public List<FolderResponse> getAllFoldersWithParent(UUID parentFolderId) {
...
+    folderRepository.save(folderEntity);
+    actionService.recordAction(Collections.singleton(folderEntity), ActionType.MOVE);
+  }
+
+  @Override
+  public void moveFolderToFolder(UUID folderId, UUID destinationFolderId) {
+    FolderEntity folderEntity =
+        folderRepository.findById(folderId).orElseThrow(FolderNotFoundException::new);
+    FolderEntity parentFolderEntity =
+        folderRepository.findById(destinationFolderId).orElseThrow(FolderNotFoundException::new);
```
ivanfhtec, line 201: why parentFolderEntity? shouldn't it be destinationFolderEntity?

backend/src/test/java/com/htecgroup/skynest/service/impl/FolderServiceImplTest.java
```diff
@@ -164,6 +164,34 @@ void getAllRootFolders() {
     verify(folderRepository, times(1)).findAllByBucketIdAndParentFolderIsNull(any());
   }
 
+  @Test
+  void moveFolderToBucket() {
```
ivanfhtec: moveFolderToBucketRoot?

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
DarkoVeljanoskiHtec: because folder is not really inside any folder here :)

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
DarkoVeljanoskiHtec: but it'll be its parent folder if you move it, I guess both are right.

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -136,6 +138,33 @@ public List<FolderResponse> getAllFoldersWithParent(UUID parentFolderId) {
...
+    if (folderEntity.getParentFolder() != null
+        && folderEntity.getParentFolder().getId() == parentFolderEntity.getId()) {
+      throw new FolderAlreadyInsideFolderException();
+    }
+    folderEntity.setParentFolder(parentFolderEntity);
```
petarstoimenovskihtec, line 204: Can we do the same logic as in the other PR? To extract the mutual logic into the one method?

backend/src/test/java/com/htecgroup/skynest/service/impl/FolderServiceImplTest.java
```diff
@@ -164,6 +164,34 @@ void getAllRootFolders() {
...
 
+  @Test
+  void moveFolderToBucketRoot() {
+    FolderEntity expectedFolderEntity = FolderEntityUtil.getFolderWithParent();
+    when(folderRepository.findById(any())).thenReturn(Optional.of(expectedFolderEntity));
+    when(folderRepository.save(any())).thenReturn(expectedFolderEntity);
+    folderService.moveFolderToRoot(UUID.randomUUID());
+    Assertions.assertNull(expectedFolderEntity.getParentFolder());
+    verify(folderRepository, times(1)).findById(any());
+    verify(folderRepository, times(1)).save(any());
```
petarstoimenovskihtec: verify with any() ? :)

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
DarkoVeljanoskiHtec: Sure

backend/src/test/java/com/htecgroup/skynest/service/impl/FolderServiceImplTest.java
DarkoVeljanoskiHtec: I'll change it same as the other PR :)

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
ivanfhtec, line 204: what if the destination folder is a child folder?
I don't think it makes sense to move a folder into a child folder
or rather, the destination must not be inside of the folder being moved

MetoMladenovski: Illusion

ivanfhtec: @DarkoVeljanoskiHtec something like this - for the destination folder, iterate over all parents until parent is not null, and check whether any parent is the folder you're moving?

ivanfhtec: like in [this PR](https://github.com/JDP2022SkyNest/SkyNest/pull/207)

DarkoVeljanoskiHtec: sounds good, just maybe yours PR should be merge first then, so we wont repeat ourselves.

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -136,6 +135,59 @@ public List<FolderResponse> getAllFoldersWithParent(UUID parentFolderId) {
...
+    if (folderEntity.getParentFolder() != null
+        && folderEntity.getParentFolder().getId() == parentFolderEntity.getId()) {
+      throw new FolderAlreadyInsideFolderException();
+    }
+  }
+
+  private void checkIfDestinationFolderIsChildFolder(
+      FolderEntity folderEntity, FolderEntity parentFolderEntity) {
+    List<FolderEntity> path = getPathToFolder(parentFolderEntity);
+    for (int i = 0; i < path.size(); i++) {
```
ivanfhtec: you could use `for (FolderEntity folder : path)` since you're not using the index

or `stream().anyMatch(...)`

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -136,6 +135,58 @@ public List<FolderResponse> getAllFoldersWithParent(UUID parentFolderId) {
...
+    checkIfDestinationFolderIsChildFolder(folderEntity, parentFolderEntity);
+    folderEntity.setParentFolder(parentFolderEntity);
+    saveMoveFolder(folderEntity);
+  }
+
+  private FolderEntity findFolderEntity(UUID folderID) {
+    return folderRepository.findById(folderID).orElseThrow(FolderNotFoundException::new);
+  }
+
+  private void checkIfFolderAlreadyInsideRoot(FolderEntity folderEntity) {
```
markobarackovhtec: Move all checks to some new folder validation service

backend/src/main/java/com/htecgroup/skynest/service/impl/FolderServiceImpl.java
```diff
@@ -137,6 +140,44 @@ public List<FolderResponse> getAllFoldersWithParent(UUID parentFolderId) {
...
+    saveMoveFolder(folderEntity);
+  }
+
+  @Override
+  public void moveFolderToFolder(UUID folderId, UUID destinationFolderId) {
+    FolderEntity folderEntity = findFolderEntity(folderId);
+    FolderEntity parentFolderEntity =
+        folderRepository.findById(destinationFolderId).orElseThrow(FolderNotFoundException::new);
+    folderValidatorService.checkIfFolderAlreadyInsideFolder(folderEntity, parentFolderEntity);
+    checkIfDestinationFolderIsChildFolder(folderEntity, parentFolderEntity);
```
markobarackovhtec: What about this check, could this also be part of folderValidationService ?

DarkoVeljanoskiHtec: This check inside is calling method from FolderServiceImpl, if I move it to FolderServiceValidator and call it from there, we will have cyclic dependency.

markobarackovhtec: Then maybe you gonna need some FolderUtil class with that method

DraganMitrasinovicHtec (approved):
Looks great

PR #188, opened 2022-07-13
SJQ2T-354: BE - Connect lambdas to buckets

MetoMladenovski:
Added a list of lambda types to bucket entity.


backend/src/main/java/com/htecgroup/skynest/model/entity/BucketEntity.java
```diff
@@ -28,4 +27,10 @@ public class BucketEntity extends ObjectEntity {
   @Column(name = "public")
   @With
   private Boolean isPublic;
+
+  @ElementCollection(targetClass = LambdaType.class)
+  @CollectionTable(name = "lambda_on_bucket", joinColumns = @JoinColumn(name = "bucket_id"))
+  @Enumerated(EnumType.STRING)
+  @Column(name = "lambda_name")
+  private List<LambdaType> lambdaTypes;
```
MarkoRadunovicHtec: Can it be simpler, for example oneToMany in annotations like in some other entities?

MetoMladenovski: LambdaType is an enum, so I can't do that. If I make a class I'm not sure how the factory is going to create lambdas. I'm looking now at some issues like if I change the name of some value in the enum it will break the database. I'm gonna look that up and see if I can find a solution.

DarkoVeljanoskiHtec (approved):
👍

backend/src/main/java/com/htecgroup/skynest/lambda/LambdaType.java
```diff
@@ -1,5 +1,16 @@
 package com.htecgroup.skynest.lambda;
 
 public enum LambdaType {
-  UPLOAD_FILE_TO_EXTERNAL_SERVICE_LAMBDA
+  // The code is used to represent the lambdas in the database and as such should not be changed.
+  UPLOAD_FILE_TO_EXTERNAL_SERVICE_LAMBDA("U_F_T_E_S_L");
+
+  private String code;
```
ivanfhtec: or text, rather than code?
this makes it look like it's the lambda's source code or something

backend/src/main/java/com/htecgroup/skynest/lambda/LambdaType.java
```diff
@@ -1,5 +1,16 @@
...
+
+  LambdaType(String code) {
+    this.code = code;
+  }
+
+  public String getCode() {
```
ivanfhtec: `@Getter`?

backend/src/main/java/com/htecgroup/skynest/lambda/converter/LambdaTypeConverter.java
```diff
@@ -0,0 +1,30 @@
+package com.htecgroup.skynest.lambda.converter;
+
+import com.htecgroup.skynest.lambda.LambdaType;
+
+import javax.persistence.AttributeConverter;
+import javax.persistence.Converter;
+import java.util.stream.Stream;
+
+@Converter(autoApply = true)
```
ivanfhtec, line 9: those annotations in BucketEntity weren't enough?

backend/src/main/java/com/htecgroup/skynest/lambda/LambdaType.java
MetoMladenovski: maybe databaseCode, since that's the value under which we are storing the enum.

backend/src/main/java/com/htecgroup/skynest/lambda/converter/LambdaTypeConverter.java
MetoMladenovski: Nope. We need this so that we can change the enum name if we want, or add new enums wherever we want without messing up the database.

backend/src/main/java/com/htecgroup/skynest/lambda/LambdaType.java
ivanfhtec: sure, or databaseString or something

or you could redefine toString for the enum, though that might cause problems later

backend/src/main/java/com/htecgroup/skynest/lambda/converter/LambdaTypeConverter.java
ivanfhtec: we did that manually for the other enums 😔

PR #191, opened 2022-07-13
SJQ2T-244: BE - Modify file contents

ivanfhtec:
* only changes file content (and size, type), not name


ivanfhtec:
forgot swagger, coming soon

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -139,6 +133,59 @@ public List<FileResponse> getAllFilesWithParent(UUID parentFolderId) {
     return asFileResponseList(allFiles);
   }
 
+  @Override
+  @Transactional(rollbackFor = Exception.class)
+  public FileResponse updateFileContent(MultipartFile multipartFile, UUID fileId) {
```
petarstoimenovskihtec: Because you have extracted this logic from the previous uploadFile method and extended with some additional update logic, I will try to separate that.
What I mean is, extract the mutual logic for saving the input stream into GridFs and corresponding logic for metadata in relation database. Then that method will be called in `uploadFile` and the same method plus the custom logic for update will be in other method that will be used in updateFileContent. 

Makes sense or I am missing something regarding this checks?

ivanfhtec: something like `saveContentAndMetadata(fileMetadata, multipartFile)`

which could then be called from uploadFileToBucket, uploadFileToFolder, and updateFileContents

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -172,6 +165,58 @@ public List<FileResponse> getAllFilesWithParent(UUID parentFolderId) {
...
+      return savedFileMetadata;
+    } catch (IOException e) {
+      log.error(e);
+      throw new FileIOException();
+    }
+  }
+
+  private void deleteFileContent(String objectId) {
+    try {
+      if (objectId == null || objectId.isEmpty()) throw new FileNotFoundException();
```
petarstoimenovskihtec: I believe there is a method from StringUtils named isEmpty that does this two checks

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -172,6 +165,58 @@ public List<FileResponse> getAllFilesWithParent(UUID parentFolderId) {
...
+      operations.delete(new Query(Criteria.where("_id").is(objectId)));
```
petarstoimenovskihtec: You can expose a method named deleteById for example in the operations service that will expect just and id (in this case objectId) and perform the Query and Criteria creation there.

That way you will hide the actual impl that is happening in operation service and expose just the interface with the corresponding method and input param.  Also you will provide the needed abstraction in object oriented programing and produce more clean and testable method.

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
ivanfhtec: `isAllBlank()` from `org.apache.commons.lang3.StringUtils`?

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
ivanfhtec: `operations` is from `org.springframework.data.mongodb.gridfs.GridFsOperations`, you mean I should extend it and add a deleteById method in that new service?

in other words, a wrapper service around GridFsOperations?

petarstoimenovskihtec: Why not? You need to hide the impl anyway, the business service doesn't need to know about the under the hood impl aka you provide abstraction

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -172,23 +157,104 @@ public List<FileResponse> getAllFilesWithParent(UUID parentFolderId) {
...
+  @Override
+  @Transactional(rollbackFor = Exception.class)
+  public FileResponse editFileContent(MultipartFile multipartFile, UUID fileId) {
+
+    FileMetadataEntity fileMetadata =
+        fileMetadataRepository.findById(fileId).orElseThrow(FileNotFoundException::new);
+
+    checkFileTypeNotChanged(fileMetadata, multipartFile.getContentType());
+
+    fileMetadata.setType(multipartFile.getContentType());
```
petarstoimenovskihtec, line 168: Since you already have a check about the file type not changed, do you actually need this here?

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
ivanfhtec: then I'll move `findFileContentById(objectId)`, `saveFileContent(objectId)`, and `deleteFileContentById(objectId)` to this new service

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
ivanfhtec: the check only checks whether the old file metadata's type is the same as the multipart file's type
then I set the new type

it's dumb, I'll change it into something readable

ivanfhtec: should be easier to read now

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -202,32 +202,54 @@ private void checkIfFileIsAlreadyInsideRoot(FileMetadataEntity fileMetadataEntit
...
       throw new FileAlreadyDeletedException();
     }
     fileMetadataEntity.delete();
     fileMetadataRepository.save(fileMetadataEntity);
   }
 
+  private FileMetadataEntity saveContentAndMetadata(
+      FileMetadataEntity fileMetadata, MultipartFile multipartFile) {
+
+    checkOnlyCreatorsCanAccessPrivateBuckets(fileMetadata);
```
markobarackovhtec, line 245: Move these checks in some new service, bucket validation or access service

ivanfhtec: there are quite a few of these checks in this service, perhaps as new PR?

DarkoVeljanoskiHtec (approved):
It's time to go

PR #192, opened 2022-07-13
SJQ2T-370: BE - Fixed modifiedOn for Objects, as well as other non-insertable fields

ivanfhtec:
* modifiedOn should now work as expected 
  * updates for files, folders, and buckets now update modifiedOn for their corresponding object

* all non-insertable fields are now also non-updatable

### DIFF
--- a/backend/src/main/resources/db/migration/V12.1__file_trigger_after_update_modified_on.sql
+++ b/backend/src/main/resources/db/migration/V12.1__file_trigger_after_update_modified_on.sql
@@ -1,0 +1,10 @@
+DELIMITER //
+CREATE TRIGGER `file_after_update_modified_on`
+  AFTER UPDATE ON `file`
+  FOR EACH ROW
+BEGIN
+  UPDATE `object`
+    SET modified_on = CURRENT_TIMESTAMP
+    WHERE object.id = NEW.id;
+END//
+DELIMITER ;
--- a/backend/src/main/java/com/htecgroup/skynest/model/entity/ActionEntity.java
+++ b/backend/src/main/java/com/htecgroup/skynest/model/entity/ActionEntity.java
@@ -22,7 +22,7 @@ public class ActionEntity implements Serializable {
 
   @Id @GeneratedValue private UUID id;
 
-  @Column(insertable = false)
+  @Column(insertable = false, updatable = false)
   private LocalDateTime performedOn;
 
   @ManyToOne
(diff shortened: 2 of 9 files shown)


backend/src/main/resources/db/migration/V12.1__file_trigger_after_update_modified_on.sql
```diff
@@ -0,0 +1,10 @@
+DELIMITER //
+CREATE TRIGGER `file_after_update_modified_on`
```
petarstoimenovskihtec, line 2: Instead of trigger, you can achieve the same with a simple sql migration. 
No need of trigger because this is one time action right?

petarstoimenovskihtec: Same applies to all other triggers

petarstoimenovskihtec: If you wonder why? The trigger lives in the DB, so potentially I can execute the trigger and change your database data

ivanfhtec: these triggers activate every time an UPDATE query updates a row, so this is not just a one time action

in this case, `file_after_update_modified_on` activates each time a row in `file` is updated

we've left the DB is in charge of maintaining `modified_on` fields in each table

but these particular tables are a problem - if we update a `FileEntity`, but don't change any part of it's base object `ObjectEntity`, hibernate will only execute an UPDATE query for the `file` table and `modified_on` won't update by itself as it's in the `object` table

petarstoimenovskihtec: hmm ok for consistency lets keep it like that. On the other hand there are JPA annotations that handle this kinds of updates.

petarstoimenovskihtec (approved):

backend/src/main/resources/db/migration/V12.1__file_trigger_after_update_modified_on.sql
ivanfhtec: I thought it would be simpler to have the DB do the heavy lifting
turns out `@CreationTimestamp` and `@UpdateTimestamp` would probably have been more straightforward

PR #193, opened 2022-07-13
SJQ2T-334 BE-Bucket must have company

MarkoRadunovicHtec:
Flyway script added to update companyId column in bucket to NOT NULL, added script to add a company to every bucket without it.


docker scripts/unix/add-company-to-all-buckets-without-company.sh
```diff
@@ -0,0 +1,9 @@
+echo "HTEC group pib:"
+echo -e "\t105793138\n"
+echo "Pib of the company you want to add buckets to"
+read -p "pib: " pib
+echo "--DB password"
+docker exec -it skynest-db mysql -u user -p --execute="USE skynest;
```
petarstoimenovskihtec: Why this is not written as a migration script with flyway?

MarkoRadunovicHtec: Because it will be only now one time if there are buckets without company already in. Any other case the company will be added to bucket upon creation because it is now refactored that every user must have a company. It can be in flyway also. It will be executed only once.

petarstoimenovskihtec: but since it is shell script, nothing stops me from executing it again right?
Every time when you need one-time script execution in database, use flyway migration because flyway it self has a mechanism that allows the scripts to be executed just once.

MarkoRadunovicHtec: Thats right. Okay, makes sense. It will be executed like flyway script just once if there is need. I'll change it to flyway script. :)

petarstoimenovskihtec (approved):

PR #201, opened 2022-07-14
SJQ2T-362 and SJQ2T-378: BE - Owner permission for each object on create

ivanfhtec:
* creating new objects will grant owner permission

* for already existing objects, granted all creators owner for their own objects through flyway


MarkoRadunovicHtec (approved):
Seems good.

PR #202, opened 2022-07-14
SJQ2T-365: BE - Endpoint for activating lambda for bucket

MetoMladenovski:
Implemented activate lambda for bucket by accepting LambdaType enum as RequestParam in the endpoint and bucket Id as PathVariable.


backend/src/main/java/com/htecgroup/skynest/controller/LambdaController.java
```diff
@@ -1,24 +1,120 @@
...
+                    @ExampleObject(
+                        value =
+                            "{\"messages\":[\"Lambda is already activated for that bucket.\"],"
+                                + " \"status\": \"409\","
+                                + " \"timestamp\": \"2022-06-07 16:18:12\"}")
+                  })
+            }),
+      })
+  @PutMapping("/bucket/{bucketId}/activate")
+  public ResponseEntity<Boolean> activateLambdaForBucket(
```
petarstoimenovskihtec: You don't need to return a boolean response, instead you can return 204 status code with no content. If I am not mistaking this was already implemented in some delete endpoint by @DraganMitrasinovicHtec

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
```diff
@@ -95,6 +97,18 @@ public List<BucketResponse> listAllDeletedBuckets() {
...
   }
 
+  @Override
+  public void activateLambda(UUID bucketId, LambdaType lambdaType) {
+    BucketEntity bucketEntity = findBucketEntityById(bucketId);
+    List<LambdaType> activatedLambdas = bucketEntity.getLambdaTypes();
+    if (activatedLambdas.contains(lambdaType)) {
+      throw new LambdaAlreadyActivated();
+    }
+    activatedLambdas.add(lambdaType);
```
petarstoimenovskihtec: maybe I would try and add a method in BucketEntity class that will add a lambda in the list..
Something like this:
```
public void addLambda(LambdaType lambdaType){
        this.getLambdaTypes().add(lambdaType);
 }
```

That way you will move the logic in the entity class and it would be reusable instead of exposing it here. Also if you disallow duplicated, consider using Set instead of List

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
```diff
@@ -95,6 +97,18 @@ public List<BucketResponse> listAllDeletedBuckets() {
...
+    BucketEntity bucketWithActivatedLambda = bucketEntity.withLambdaTypes(activatedLambdas);
+    bucketRepository.save(bucketWithActivatedLambda);
```
petarstoimenovskihtec: If I am not mistaking if the relation is done ok, the save method is obsolete if you add the lambdaType the way I describe in the previous comment

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
MetoMladenovski: I considered the two of these. The first one, I don't know why I didn't do it..
And using a Set might bring me trouble down the road since it doesn't really have a get(), so I figured sticking with a List is better. 
Edit: But if I really think about it, i think I don't need a get(). I will try with Set.

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
MetoMladenovski: I think so too, yes.
But the save() method is needed to save the changes in the database. I mean can we ever save anything in database without it? I'm confused a little bit.

backend/src/main/java/com/htecgroup/skynest/model/entity/BucketEntity.java
```diff
@@ -31,5 +31,10 @@ public class BucketEntity extends ObjectEntity {
   @ElementCollection(targetClass = LambdaType.class)
   @CollectionTable(name = "lambda_on_bucket", joinColumns = @JoinColumn(name = "bucket_id"))
   @Column(name = "lambda_name")
-  private List<LambdaType> lambdaTypes;
+  @With
```
petarstoimenovskihtec: You don't need the @With annotation now, right?

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
```diff
@@ -98,6 +99,13 @@ public List<BucketResponse> listAllDeletedBuckets() {
         .collect(Collectors.toList());
   }
 
+  @Override
+  public void activateLambda(UUID bucketId, LambdaType lambdaType) {
+    BucketEntity bucketEntity = findBucketEntityById(bucketId);
+    bucketEntity.addLambda(lambdaType);
+    bucketRepository.save(bucketEntity);
```
petarstoimenovskihtec, line 107: Did you try the code without the save method?

backend/src/main/java/com/htecgroup/skynest/model/entity/BucketEntity.java
MetoMladenovski: Yes, I removed it but didn't push the code.

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
MetoMladenovski: I tried it, but it's not saving the added lambda in the database.

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
petarstoimenovskihtec: did you try it?

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
petarstoimenovskihtec: ok, maybe a transaction manager is needed, but leave it like that

petarstoimenovskihtec (approved):

PR #203, opened 2022-07-15
work in progress

MilenaHtec:


frontend/src/Components/axios/CustomAxios.jsx
```diff
@@ -1,34 +1,48 @@
...
+import AxiosInstance from "./AxiosInstance";
+import { getRefreshToken } from "../ReusableComponents/ReusableFunctions";
 
-const CustomAxios = axios.create({});
-CustomAxios.interceptors.request.use(
-   (request) => {
-      request.headers["accessToken"] = localStorage.getItem("accessToken");
-      request.headers["refresh-token"] = localStorage.getItem("refresh-token");
-      return request;
+AxiosInstance.interceptors.request.use(
```
marija-antic, line 4: I think you should check your access token i request not in response intercept. Here check your token and expirattionn time and see if it is expired. if it is pull refresh token, set new access token etc.


here is some real world example:

`  axios.interceptors.request.use(async (config: any) => {
    const accessToken: string | null = getAccessToken();
    const { exp: tokenExparation } = decodeJwtToken(accessToken);

    try {
      if (
        accessToken &&
        tokenExparation <= Math.trunc((Date.now() + 30000) / 1000)
      ) {
        const response = await axios.post(
          `${api}/auth/token?grant_type=refresh_token`,
        );
        config.headers.Authorization = `Bearer ${response.data.accessToken}`;
        setAccessToken(response.data.accessToken);
      }
    } catch {
      removeAccessToken();
      history.push({
        pathname: '/login',
      });
    }

    return config;
  });`

PR #205, opened 2022-07-15
SJQ2T-382: BE - Something is deleted if it's contained in something deleted

ivanfhtec:
* added isDeleted for files and folders
* added restore method for future use
* fixed tests to comply


DarkoVeljanoskiHtec (approved):
Looks legit.

PR #207, opened 2022-07-15
SJQ2T-383: BE - Folder contents now also includes path

ivanfhtec:
* get folder contents now also includes path to folder

I'll update docs if this looks good


DarkoVeljanoskiHtec (approved):
Looks good

PR #209, opened 2022-07-15
SJQ2T-387: BE - Restore folder, small changes to delete folder

ivanfhtec:
* a folder can be restored if it is deleted, but its parent is not

* folders are considered to be deleted if deletedOn is not null, or if they are contained inside of something that’s deleted

* also changed delete folder to reflect this definition of delete - deleted folders cannot be deleted

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
+++ b/backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
@@ -233,7 +233,79 @@ public class FolderController {
     folderService.removeFolder(folderId);
   }
 
-  @Operation(summary = "Get folder contents")
+  @Operation(summary = "Restore Folder")
+  @ApiResponses(
+      value = {
+        @ApiResponse(
+            responseCode = "200",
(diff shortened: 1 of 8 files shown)


ivanfhtec:
I've tested this manually and it seems to work properly

I'll write unit tests if this seems OK

backend/src/main/java/com/htecgroup/skynest/controller/FolderController.java
```diff
@@ -233,7 +233,79 @@ public void removeFolder(@PathVariable UUID folderId) {
     folderService.removeFolder(folderId);
   }
 
-  @Operation(summary = "Get folder contents")
+  @Operation(summary = "Restore Folder")
+  @ApiResponses(
+      value = {
+        @ApiResponse(
+            responseCode = "200",
+            description = "Folder successfully removed",
```
MarkoRadunovicHtec, line 241: Folder successfully restored?

markobarackovhtec (approved):

PR #214, opened 2022-07-18
SJQ2T-366: BE - Endpoint for deactivating lambda for bucket

MetoMladenovski:
The same logic as activate lambda.


backend/src/main/java/com/htecgroup/skynest/controller/LambdaController.java
```diff
@@ -67,6 +67,51 @@ public List<LambdaType> getAllLambdas() {
...
+                                + " \"timestamp\": \"2022-06-07 16:18:12\"}")
+                  })
+            }),
+      })
+  @PutMapping("/bucket/{bucketId}/deactivate")
+  @ResponseStatus(HttpStatus.NO_CONTENT)
+  public void deactivateLambdaForBucket(
+      @PathVariable UUID bucketId, @RequestParam LambdaType lambda) {
+    bucketService.deactivateLambda(bucketId, lambda);
+    log.info("Deactivated lambda {} for bucket {}", lambda.toString(), bucketId.toString());
```
MarkoRadunovicHtec: Maybe to log also which owner deactivated labmda..

MetoMladenovski: That would be nice, log the user that activated/deactivated the lambda. Sure

markobarackovhtec (approved):

PR #224, opened 2022-07-19
SJQ2T-385: BE - Create lambda upload file to external service

MetoMladenovski:
The ACCESS_TOKEN is currently defined directly in the creation of the bean. I had an issue where the token expired after a while. If the token doesn't expire and everything is okay, then it can go with the other .env variables and keep it there.


backend/src/main/java/com/htecgroup/skynest/SkyNestApplication.java
```diff
@@ -40,4 +42,12 @@ public DriverConfigLoaderBuilderCustomizer defaultProfile() {
             .withString(DefaultDriverOption.METADATA_SCHEMA_REQUEST_TIMEOUT, "20 seconds")
             .build();
   }
+
+  @Bean("dropboxClient")
+  public DbxClientV2 dropboxClient() {
+    String ACCESS_TOKEN =
+        "sl.BLvQJNuOz5yfPt5TSENtKsPX53OUioAOpcZez0-cvC_zjDOv-wxnXhK6s6eQ__9J3FVUuAnkDzjVaZDQo6nz9Jrr991MWDmChdWLIn8gdZkPOhwgkjnlZAaGz9tKUpG0Ea7DrFY";
```
ivanfhtec: > The ACCESS_TOKEN is currently defined directly in the creation of the bean. I had an issue where the token expired after a while. If the token doesn't expire and everything is okay, then it can go with the other .env variables and keep it there.

won't it expire anyway? what difference would it make if it were an env var?

backend/src/main/java/com/htecgroup/skynest/lambda/impl/UploadFileToExternalServiceLambda.java
```diff
@@ -1,16 +1,49 @@
...
-  public Boolean doLambdaFunction(String objects) {
+  public Boolean doLambdaFunction(UploadFileToExternalServiceEvent event) {
+    try {
+      MultipartFile file = event.getFileToUpload();
+      ByteArrayInputStream inputStream = new ByteArrayInputStream(file.getBytes());
+      String fileName = event.getFileToUpload().getOriginalFilename();
+      Metadata uploadMetaData =
+          dropboxClient
+              .files()
+              .uploadBuilder("/" + event.getBucketName() + "/" + fileName)
```
ivanfhtec, line 39: should probably use `String.format()`

also, shouldn't this contain bucket ids, and not names? names aren't unique

backend/src/main/java/com/htecgroup/skynest/SkyNestApplication.java
MetoMladenovski: the only difference it makes is that I lose time if I put it in the env and it expires. Anyhow this is all changed now. It's not gonna be this way

backend/src/main/java/com/htecgroup/skynest/lambda/impl/UploadFileToExternalServiceLambda.java
MetoMladenovski: I didn't like it with ids, it's way more useless seeing that UUID there for the bucket..

ivanfhtec: it is, but there could be two buckets with the same name
you could include both somehow

MetoMladenovski: The files just go in the same bucket on dropbox. We don't have path on dropbox anyhow. It's just the buckets.

DarkoVeljanoskiHtec (approved):
LAMBDA

PR #236, opened 2022-07-21
SJQ2T-324: BE Permissions For Edit Bucket

DarkoVeljanoskiHtec:


backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
```diff
@@ -151,6 +148,7 @@ public void restoreBucket(UUID uuid) {
   @Override
   public BucketResponse editBucket(
       BucketEditRequest bucketEditRequest, @Valid @CurrentUserCanEditBucket UUID uuid) {
+    permissionService.currentUserHasPermissionForBucket(uuid, AccessType.EDIT);
```
DraganMitrasinovicHtec: Did you test this, I think this annotation is old and it only checks for ownership

ivanfhtec: the method is old, but should work

ivanfhtec: not sure the annotation is necessary

DarkoVeljanoskiHtec: should i remove it ?

DraganMitrasinovicHtec: Yeah, just remove the annotation

ivanfhtec (approved):

PR #242, opened 2022-07-21
SJQ2T - 447: New migrations, and some register documentation

DraganMitrasinovicHtec:
Made company name unique, companyId in user table is not null, and created a default admin user for company HTEC


MarkoRadunovicHtec (approved):
Legit. Needed fix.

PR #252, opened 2022-07-25
SJQ2T-375 BE - Untag object endpoint

MarkoRadunovicHtec:
Endpoint for untag object implemented, tests added.

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/service/impl/TagServiceImpl.java
+++ b/backend/src/main/java/com/htecgroup/skynest/service/impl/TagServiceImpl.java
@@ -2,10 +2,7 @@ package com.htecgroup.skynest.service.impl;
 
 import com.htecgroup.skynest.exception.company.CompanyNotFoundException;
 import com.htecgroup.skynest.exception.object.ObjectNotFoundException;
-import com.htecgroup.skynest.exception.tag.TagAlreadyExistsException;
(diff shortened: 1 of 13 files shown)


backend/src/main/java/com/htecgroup/skynest/service/impl/TagServiceImpl.java
```diff
@@ -119,4 +116,34 @@ public List<TagResponse> getTagsForObject(UUID objectId) {
...
+      ObjectEntity objectEntity =
+              objectRepository.findById(objectId).orElseThrow(ObjectNotFoundException::new);
+
+      ObjectToTagKey key = new ObjectToTagKey(tagId, objectId);
+      ObjectToTagEntity objectToTagEntity = new ObjectToTagEntity(key, tagEntity, objectEntity);
+
+      if (!objectToTagRepository.existsById(key)) {
+        throw new TagOnObjectNotFound();
+      }
+      if (!tagEntity.getCompany().equals(objectEntity.getCreatedBy().getCompany())) {
```
ivanfhtec: not sure if `CompanyEntity` has an `equals()` method
you could compare by ids?

MarkoRadunovicHtec: Will check. Will also remove checks from here and place them in validation service.

DraganMitrasinovicHtec (approved):
No comment

PR #253, opened 2022-07-25
SJQ2T-409: BE - Permission check for getFileDetails

ivanfhtec:
* also added recursive permission check for folders


DarkoVeljanoskiHtec (approved):
Looks good

PR #255, opened 2022-07-25
SJQ2T-386: BE - Restore file endpoint

DraganMitrasinovicHtec:
Endpoint for restoring files, I only allowed restoring if the File is deleted and none of it's parents (including the bucket) are not.

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/model/entity/FileMetadataEntity.java
+++ b/backend/src/main/java/com/htecgroup/skynest/model/entity/FileMetadataEntity.java
@@ -37,6 +37,10 @@ public class FileMetadataEntity extends ObjectEntity {
         || (parentFolder == null ? bucket.isDeleted() : parentFolder.isDeleted());
   }
 
+  public boolean someParentIsDeleted() {
+    return parentFolder == null ? bucket.isDeleted() : parentFolder.isDeleted();
+  }
+
   public void moveToRoot(FileMetadataEntity fileMetadataEntity) {
     fileMetadataEntity.setParentFolder(null);
   }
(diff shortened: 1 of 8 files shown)


backend/src/main/java/com/htecgroup/skynest/model/entity/FileMetadataEntity.java
```diff
@@ -37,6 +37,10 @@ public boolean isDeleted() {
         || (parentFolder == null ? bucket.isDeleted() : parentFolder.isDeleted());
   }
 
+  public boolean someParentIsDeleted() {
```
ivanfhtec, line 39: you could call this in `isDeleted()` above

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -243,6 +243,34 @@ public void deleteFile(UUID fileId) {
...
 
+  @Override
+  @RecordAction(objectId = "[0].toString()", actionType = ActionType.RESTORE)
+  public FileResponse restoreFile(UUID fileId) {
+    LoggedUserDto currentUser = currentUserService.getLoggedUser();
+
+    FileMetadataEntity fileMetadataEntity =
+        fileMetadataRepository.findById(fileId).orElseThrow(FileNotFoundException::new);
+
+    if (!((ObjectEntity) fileMetadataEntity).isDeleted()) {
```
ivanfhtec: this cast shouldn't be necessary

backend/src/main/java/com/htecgroup/skynest/model/entity/FileMetadataEntity.java
ivanfhtec: we could do the same for folders

DraganMitrasinovicHtec: Added this

backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
DraganMitrasinovicHtec: Got my logic mixed up, ur right

DarkoVeljanoskiHtec (approved):
shinzou wo sasageyo !

PR #257, opened 2022-07-25
SJQ2T-322 BE Permissions For Get All Buckets

DarkoVeljanoskiHtec:


backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
```diff
@@ -103,10 +104,21 @@ public void deactivateLambda(UUID bucketId, LambdaType lambda) {
 
   @Override
   public List<BucketResponse> listAllBuckets() {
+
     List<BucketEntity> entityList =
         (List<BucketEntity>) bucketRepository.findAllByOrderByNameAscCreatedOnDesc();
+    List<BucketEntity> newEntityList = new ArrayList<BucketEntity>();
+
+    for (BucketEntity bucket : entityList) {
```
DraganMitrasinovicHtec: Use streams for this, filter method is perfect

DarkoVeljanoskiHtec: you have an idea how should it look ? because currentUserHasPermissionForBucket throws an exception for those who you don't have permissions on, thats why I had to do it with try catch block.

ivanfhtec: the filter predicate could be something like this

```
try {
  permissionService.currentUserHasPermissionForBucket(bucket.getId(), AccessType.VIEW);
  newEntityList.add(bucket);
} catch (BucketAccessDeniedException b) {
  return false;
}
return true;
```

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
```diff
@@ -103,8 +104,10 @@ public void deactivateLambda(UUID bucketId, LambdaType lambda) {
 
   @Override
   public List<BucketResponse> listAllBuckets() {
+
     List<BucketEntity> entityList =
         (List<BucketEntity>) bucketRepository.findAllByOrderByNameAscCreatedOnDesc();
+    entityList.stream().filter(b -> doesCurrentUserHaveViewPermissionOnBucket(b));
```
ivanfhtec: I think you could also do
```
this::doesCurrentUserHaveViewPermissionOnBucket
```
instead of
```
b -> doesCurrentUserHaveViewPermissionOnBucket(b)
```

not sure which one looks cleaner

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
```diff
@@ -114,6 +117,15 @@ public List<BucketResponse> listAllBuckets() {
         .collect(Collectors.toList());
   }
 
+  private boolean doesCurrentUserHaveViewPermissionOnBucket(BucketEntity bucket) {
+    try {
+      permissionService.currentUserHasPermissionForBucket(bucket.getId(), AccessType.VIEW);
```
ivanfhtec, line 122: this method only checks the bucket for permissions

in order for this to work, when we grant permissions for a file or folder, we need to grant view permission for everything on the file/folder's path, bucket included

or rather, for everything on the path where the user doesn't already have some kind of permission

DraganMitrasinovicHtec: Yeah, that was the plan

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
DarkoVeljanoskiHtec: I guess the first one

DraganMitrasinovicHtec (approved):
Tatakae!

PR #262, opened 2022-07-26
SJQ2T-411: BE - Permissions for uploadFileToBucket

ivanfhtec:
* upload to folder keeps old behaviour


DraganMitrasinovicHtec (approved):
If it works, it works

PR #268, opened 2022-07-26
SJQ2T-416: BE Permissions For Edit File Content

DarkoVeljanoskiHtec:


backend/src/main/java/com/htecgroup/skynest/service/impl/FileServiceImpl.java
```diff
@@ -165,6 +165,9 @@ public FileResponse editFileContent(MultipartFile multipartFile, UUID fileId) {
     FileMetadataEntity fileMetadata =
         fileMetadataRepository.findById(fileId).orElseThrow(FileNotFoundException::new);
 
+    checkIfDeleted(fileMetadata);
```
DraganMitrasinovicHtec, line 176: Do we need a public check here?

DarkoVeljanoskiHtec: Yes we need

ivanfhtec (approved):

PR #285, opened 2022-07-26
SJQ2T-422: BE - Get all folder permissions

MetoMladenovski:


backend/src/main/java/com/htecgroup/skynest/service/impl/PermissionServiceImpl.java
```diff
@@ -198,10 +197,6 @@ public List<PermissionResponse> getAllBucketPermission(UUID bucketId) {
         .collect(Collectors.toList());
   }
 
-  private void checkIfBucketExist(UUID bucketId) {
```
DraganMitrasinovicHtec, line 201: Why are we removing this again?

backend/src/main/java/com/htecgroup/skynest/service/impl/PermissionServiceImpl.java
```diff
@@ -237,6 +230,26 @@ public void revokePermission(UUID bucketId, UUID userId) {
...
   }
 
+  @Override
+  public List<PermissionResponse> getAllFolderPermission(UUID folderId) {
+    FolderEntity folder =
+        folderRepository.findById(folderId).orElseThrow(FolderNotFoundException::new);
+    checkIfFolderIsDeleted(folder);
+    currentUserHasPermissionForFolder(folder, AccessType.OWNER);
+    List<UserObjectAccessEntity> entityList = permissionRepository.findAllByObjectId(folderId);
+
```
DraganMitrasinovicHtec, line 249: Maybe copy logging from Ivan so we keep it consistent :D

DarkoVeljanoskiHtec (approved):

PR #287, opened 2022-07-27
SJQ2T 428: BE Revoke Files Permissions

DarkoVeljanoskiHtec:


backend/src/main/java/com/htecgroup/skynest/service/impl/PermissionServiceImpl.java
```diff
@@ -237,6 +237,21 @@ public void revokePermission(UUID bucketId, UUID userId) {
...
   }
 
+  @Override
+  public void revokeFilePermission(UUID fileId, UUID userId) throws FileNotFoundException {
+    FileMetadataEntity fileMetadataEntity =
+        fileRepository.findById(fileId).orElseThrow(FileNotFoundException::new);
+    checkIfBucketIsDeleted(fileMetadataEntity.getBucket().getId());
+    currentUserHasPermissionForFile(fileMetadataEntity, AccessType.OWNER);
+    currentUserHasPermissionForFolder(fileMetadataEntity.getParentFolder(), AccessType.OWNER);
+    currentUserHasPermissionForBucket(fileMetadataEntity.getBucket().getId(), AccessType.OWNER);
```
ivanfhtec: I don't think these last two checks are necessary
`currentUserHasPermissionForFile()` should be enough

backend/src/main/java/com/htecgroup/skynest/service/impl/PermissionServiceImpl.java
```diff
@@ -237,6 +237,21 @@ public void revokePermission(UUID bucketId, UUID userId) {
...
+
+    UserEntity user = userRepository.findById(userId).orElseThrow(UserNotFoundException::new);
+    UserObjectAccessEntity permission =
+        permissionRepository.findByObjectIdAndGrantedTo(fileId, user);
```
ivanfhtec: this doesn't return an optional? should we refactor?
if this returns a null, we should throw an error?

DarkoVeljanoskiHtec: It does return Optional, but it was asking  for throw in method signature as well, I can't say why.

backend/src/main/java/com/htecgroup/skynest/service/PermissionService.java
```diff
@@ -29,6 +30,8 @@ public interface PermissionService {
 
   void revokePermission(UUID bucketId, String userEmail);
 
+  void revokeFilePermission(UUID fileId, String email) throws FileNotFoundException;
```
MarkoRadunovicHtec: Why is throws declaration there?

DarkoVeljanoskiHtec: Intelij is asking for it, I don't really understand why.

backend/src/main/java/com/htecgroup/skynest/service/impl/BucketServiceImpl.java
```diff
@@ -124,7 +124,9 @@ public List<BucketResponse> listAllBuckets() {
 
     List<BucketEntity> entityList =
         (List<BucketEntity>) bucketRepository.findAllByOrderByNameAscCreatedOnDesc();
-    entityList.stream().filter(this::doesCurrentUserHaveViewPermissionOnBucket);
+    entityList.stream()
+        .filter(this::doesCurrentUserHaveViewPermissionOnBucket)
+        .collect(Collectors.toList());
```
ivanfhtec, line 129: shouldn't this be
```
entityList = entityList.stream()
        .filter(this::doesCurrentUserHaveViewPermissionOnBucket)
```

backend/src/main/java/com/htecgroup/skynest/service/impl/PermissionServiceImpl.java
```diff
@@ -228,13 +233,42 @@ public void revokePermission(UUID bucketId, String userEmail) {
...
     currentUserHasPermissionForBucket(bucketId, AccessType.OWNER);
 
-    UserEntity user = userRepository.findUserByEmail(userEmail).orElseThrow(UserNotFoundException::new);
+    UserEntity user =
+        userRepository.findUserByEmail(userEmail).orElseThrow(UserNotFoundException::new);
+    checkIfTargetUserIsTheCurrentUser(user);
+    UserObjectAccessEntity permission =
+        permissionRepository
+            .findByObjectIdAndGrantedTo(bucketId, user)
+            .orElseThrow(BucketNotFoundException::new);
```
ivanfhtec, line 242: should be PermissionDoesNotExist

backend/src/main/java/com/htecgroup/skynest/service/impl/PermissionServiceImpl.java
```diff
@@ -215,8 +218,10 @@ public PermissionResponse editPermission(
       PermissionEditRequest permissionEditRequest, UUID bucketId) {
     checkIfBucketIsDeleted(bucketId);
     currentUserHasPermissionForBucket(bucketId, AccessType.EDIT);
-    UserObjectAccessEntity userObjectAccessEntity = permissionRepository.findByObjectId(bucketId);
+    UserObjectAccessEntity userObjectAccessEntity =
+        permissionRepository.findByObjectId(bucketId).orElseThrow(BucketNotFoundException::new);
```
ivanfhtec, line 222: should be PermissionDoesNotExist

backend/src/main/java/com/htecgroup/skynest/service/impl/PermissionServiceImpl.java
```diff
@@ -215,8 +218,10 @@ public PermissionResponse editPermission(
       PermissionEditRequest permissionEditRequest, UUID bucketId) {
     checkIfBucketIsDeleted(bucketId);
     currentUserHasPermissionForBucket(bucketId, AccessType.EDIT);
```
ivanfhtec, line 220: OWNER?

MetoMladenovski (approved):

backend/src/main/java/com/htecgroup/skynest/service/impl/PermissionServiceImpl.java
```diff
@@ -215,8 +218,10 @@ public PermissionResponse editPermission(
...
-    UserObjectAccessEntity userObjectAccessEntity = permissionRepository.findByObjectId(bucketId);
+    UserObjectAccessEntity userObjectAccessEntity =
+        permissionRepository.findByObjectId(bucketId).orElseThrow(BucketNotFoundException::new);
     UserEntity targetUser = findTargetUserForEdit(permissionEditRequest);
+    checkIfTargetUserIsTheCurrentUser(targetUser);
```
ivanfhtec, line 224: this looks like the target should be the current user
how about `checkCurrentUserNotTargetUser`?

backend/src/main/java/com/htecgroup/skynest/service/impl/PermissionServiceImpl.java
```diff
@@ -228,13 +233,42 @@ public void revokePermission(UUID bucketId, String userEmail) {
...
+    checkIfFileIsDeleted(fileMetadataEntity);
+    currentUserHasPermissionForFile(fileMetadataEntity, AccessType.OWNER);
+
+    UserEntity user = userRepository.findUserByEmail(email).orElseThrow(UserNotFoundException::new);
+    checkIfTargetUserIsTheCurrentUser(user);
     UserObjectAccessEntity permission =
-        permissionRepository.findByObjectIdAndGrantedTo(bucketId, user);
+        permissionRepository
+            .findByObjectIdAndGrantedTo(fileId, user)
+            .orElseThrow(FileNotFoundException::new);
```
ivanfhtec, line 260: PermissionDoesNotExist

backend/src/main/java/com/htecgroup/skynest/service/impl/PermissionServiceImpl.java
```diff
@@ -228,13 +233,42 @@ public void revokePermission(UUID bucketId, String userEmail) {
...
     checkIfPermissionExist(permission);
```
ivanfhtec, line 261: doesn't the line above check for this already?

backend/src/main/java/com/htecgroup/skynest/service/impl/PermissionServiceImpl.java
```diff
@@ -228,13 +233,42 @@ public void revokePermission(UUID bucketId, String userEmail) {
...
+    checkIfPermissionExist(permission);
+    permissionRepository.delete(permission);
+  }
+
+  @Override
+  public void revokeFilePermission(UUID fileId, String email) {
+    FileMetadataEntity fileMetadataEntity =
+        fileRepository.findById(fileId).orElseThrow(FileNotFoundException::new);
+    checkIfBucketIsDeleted(fileMetadataEntity.getBucket().getId());
+    checkIfFileIsDeleted(fileMetadataEntity);
```
ivanfhtec, line 252: if the bucket is deleted, `checkIfFileIsDeleted` would return true
I don't think you need `checkIfBucketIsDeleted`

PR #295, opened 2022-07-27
SJQ2T-456 BE - Sort all tags

MarkoRadunovicHtec:
Will add index for tag name for database


DraganMitrasinovicHtec (approved):
Seems sufficiently sorted

PR #303, opened 2022-07-27
SJQ2T-360: BE - Edit file permissions

ivanfhtec:
* also renamed bucket modify/revoke methods

### DIFF
--- a/backend/src/main/java/com/htecgroup/skynest/controller/PermissionController.java
+++ b/backend/src/main/java/com/htecgroup/skynest/controller/PermissionController.java
@@ -150,7 +150,7 @@ public class PermissionController {
             })
       })
   @PostMapping("/bucket")
-  public ResponseEntity<PermissionResponse> grantPermissions(
+  public ResponseEntity<PermissionResponse> grantPermissionForBucket(
       @Valid @RequestBody PermissionGrantRequest permissionGrantRequest) {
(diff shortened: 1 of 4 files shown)


backend/src/main/java/com/htecgroup/skynest/controller/PermissionController.java
```diff
@@ -655,4 +656,66 @@ public ResponseEntity<PermissionResponse> grantPermissionToFile(
...
+                        value =
+                            "{\"messages\":[\"File not found\"],"
+                                + " \"status\": \"404\","
+                                + " \"timestamp\": \"2022-06-07 16:18:12\"}")
+                  })
+            }),
+      })
+  @PutMapping("/file/{fileId}")
+  public ResponseEntity<PermissionResponse> editPermissionForFile(
+      @Valid @RequestBody PermissionEditRequest permissionEditRequest, @PathVariable UUID fileId) {
```
DraganMitrasinovicHtec, line 842: Should we maybe add objectId to PermissionEditRequest?

ivanfhtec: I suspect this would break more than one test

DraganMitrasinovicHtec (approved):
Too long, didn't read

PR #304, opened 2022-07-27
SJQ2T-420: BE Revoke Folder Permissions

DarkoVeljanoskiHtec:


DraganMitrasinovicHtec (approved):
Too short, probably not important

PR #318, opened 2022-07-28
SJQ2T-459: BE - Create lambda "Send stats for bucket to email lambda"

MetoMladenovski:


backend/src/main/java/com/htecgroup/skynest/lambda/impl/SendBucketStatsToEmailLambda.java
```diff
@@ -0,0 +1,78 @@
...
+
+  private List<FileStatsEmailResponse> getFileStatsEmailResponses(
+      List<FileResponse> fileResponses) {
+    return fileResponses.stream()
+        .map(
+            fileResponse -> {
+              Integer downloads =
+                  actionService
+                      .getActionsWithTypeForObject(ActionType.DOWNLOAD, fileResponse.getId(), true)
+                      .size();
```
ivanfhtec, line 59: you could add a repo method `countActionsWithTypeForObject(...)` if you only need the count

backend/src/main/java/com/htecgroup/skynest/lambda/impl/SendBucketStatsToEmailLambda.java
```diff
@@ -0,0 +1,78 @@
...
+              Integer edits =
+                  actionService
+                      .getActionsWithTypeForObject(ActionType.EDIT, fileResponse.getId(), true)
+                      .size();
+              Integer moves =
+                  actionService
+                      .getActionsWithTypeForObject(ActionType.MOVE, fileResponse.getId(), true)
+                      .size();
+              Double size = Double.parseDouble(fileResponse.getSize()) / 1000000;
```
ivanfhtec: should this be 1024 * 1024, i.e. MiB?
1000 * 1000 is MB
just checking

backend/src/main/java/com/htecgroup/skynest/lambda/impl/SendBucketStatsToEmailLambda.java
ivanfhtec, line 68: since you only need the count and not the actions themselves, this should probably use count in the action repo
which would require a new method in the action service

but this works, so it's fine

backend/src/main/java/com/htecgroup/skynest/model/response/FileStatsEmailResponse.java
```diff
@@ -0,0 +1,17 @@
...
+
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.Setter;
+
+@AllArgsConstructor
+@Getter
+@Setter
+public class FileStatsEmailResponse {
+  private String fileName;
```
ivanfhtec, line 11: should this have the path as well?

DraganMitrasinovicHtec (approved):
Approve first, ask questions later

DarkoVeljanoskiHtec (approved):

ivanfhtec (approved):

PR #327, opened 2022-08-01
BE - HOTFIXES

ivanfhtec:
* folder path now contains target folder

* user must have owner to grant bucket permission


DarkoVeljanoskiHtec (approved):

ivanfhtec:
* also unintentionally fixed moving a folder into itself

